PR #6, opened 2024-10-08
[Agent] Start in INITIALIZED phase

tryuan99:
All agents start in the `INITIALIZED` phase.  During the `INITIALIZED` phase, the agent is subject to no forces (i.e., remains stationary).  During the `READY` stage, the agent is subject to gravity and drag with zero acceleration input.

This distinction is necessary later when a submunition is dispensed (in the `READY` mode) but has not fired its booster yet.

### DIFF
--- a/Assets/Scripts/Agent.cs
+++ b/Assets/Scripts/Agent.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 using UnityEngine;
 
 public abstract class Agent : MonoBehaviour {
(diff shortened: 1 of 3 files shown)


daniellovell:
Seems reasonable, a bit worried that it will produce strange behaviors on threats, but seems to pass tests. I'll test further.

tryuan99:
In `SimManager`, the threats are automatically moved into the `MIDCOURSE` phase, so for them, they don't care about `INITIALIZED` or `READY`.

PR #7, opened 2024-10-08
[Agent] Introduce maximum forward acceleration

tryuan99:
Currently, agents have a maximum reference acceleration, which refers to the normal acceleration.  This PR introduces a maximum forward acceleration, which is only used by threats currently, that defines the maximum acceleration at which an agent can accelerate or decelerate along its roll axis.


daniellovell:
Needed to resolve merge conflicts

PR #9, opened 2024-10-09
[Tracker] Add target model to track targets

tryuan99:
This PR adds a dummy agent as a simple target model when a target is assigned.  The target model simply retains the sensed position and velocity of the target, and the acceleration input at each time step is calculated with respect to the target model.  At each sense period, the target model's state is updated from the actual target.

This PR was tested by setting the sensor frequency of the micromissiles to 0.1 Hz.
Without the target model, the interceptors get lost:
before

With the target model, the interceptors hit their targets (due to zero target acceleration).
after

This PR also renames `DummyThreatTarget` to `DummyAgent` and removes the cube from the prefab.


daniellovell:
Awesome!! this will probably allow us to set the intercept distance from 10m -> 1m

daniellovell (approved):
LGTM

PR #10, opened 2024-10-10
[Controls] Add augmented proportional navigation

tryuan99:
This PR changes the interceptors to use augmented proportional navigation.  (Dummy agents can't accelerate, so augmented PN isn't enabled for threats).

Does augmented PN help?  Kinda.  Empirically, we see that the interceptors are leading the threats.  We simulated the engagement and modified the threats to be accelerating at 25g towards the asset.  However, we still miss quite a few targets (not sure if this is due to the simulation accuracy from the increased threat acceleration).

Without augmented PN:
without

With augmented PN:
with
with_hit

### DIFF
--- a/Assets/Scripts/Agent.cs
+++ b/Assets/Scripts/Agent.cs
@@ -26,6 +26,7 @@ public abstract class Agent : MonoBehaviour {
 
   [SerializeField]
   protected Agent _target;
+  [SerializeField]
   protected Agent _targetModel;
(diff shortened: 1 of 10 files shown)


daniellovell:
Would be nice to see if the logging + Python analysis script can be used to show numerically that APN is better than PN

daniellovell:
Can APN be separated from PN so we can make it configurable?

PR #16, opened 2024-10-16
Reassign interceptors during termination

tryuan99:
This PR unassigns the targets and re-assigns a threat's interceptors during agent termination.

### DIFF
--- a/Assets/Scripts/Agent.cs
+++ b/Assets/Scripts/Agent.cs
@@ -115,14 +115,10 @@ public abstract class Agent : MonoBehaviour {
     return _targetModel;
   }
 
-  public void CheckTargetHit() {
-    if (HasAssignedTarget() && _target.IsHit()) {
-      UnassignTarget();
-    }
-  }
-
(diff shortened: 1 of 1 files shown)


tryuan99:
@daniellovell All agents need to go through the `TerminateAgent()` function, but somehow, some interceptors aren't being reassigned correctly (even through their targets have been terminated).  Do you know why this might be the case?

Interceptor reassignment

daniellovell:
I think the trouble is coming from wrong level of abstraction. Should be done in the IADS.cs, which has a `IADS::MarkThreatDestroyed` method. If its implemented in IADS, we can also make it configurable in a "limited visibility" environment

daniellovell (changes requested):
I think the trouble is coming from wrong level of abstraction. Should be done in the `IADS.cs`, which has a `IADS::MarkThreatDestroyed` method. If its implemented in IADS, we can also make it configurable in a "limited visibility" environment

PR #21, opened 2025-01-06
[Planning] Add 2D interpolator for calculating the optimal launch angle

tryuan99:
This PR adds a 2D interpolator class that, for now, performs nearest neighbor interpolation using the data points' 2D coordinates.  The remaining values are then interpolated based on the nearest data point.

This 2D interpolator is then used by the `LaunchAngleInterpolator` that, given the target distance and altitude, returns the optimal launch angle and time to reach the target position.

PR #22, opened 2025-01-08
[Prediction/Planning] Add launch planner

tryuan99:
This PR adds a launch planner that integrates a predictor for the target trajectory and a launch angle planner.

**Predictor**
This PR adds an `IPredictor` interface that defines the API for predicting the target trajectory given its current state (and perhaps past states).  For now, the predictor is a simple linear extrapolator that simply extrapolates the trajectory of the target based on its current position and velocity.

**Launch angle planner**
This PR adds an `ILaunchAnglePlanner` interface for any planner that determines the optimal launch angle and time-to-intercept given the target position.  This interface now wraps the 2D interpolator from #21.

**Launch planner**
The launch planner `ILaunchPlanner` interface defines the API for planning whether to launch an interceptor against a target given a predictor for the target trajectory and a launch angle planner.  The planner essentially has to find the intersection between the optimal interceptor trajectory and the target trajectory, such that the time-to-intercept is the same for both the interceptor and the target.

For now, I implemented a simple iterative launch planner that performs a 2-step iterative process to find this intersection point:

1. _Time-to-intercept estimation_: The algorithm determines the time it takes the interceptor to reach the target at the predicted intercept position, which is initialized to the target's current position.
2. _Intercept position estimation_: The algorithm predicts the target position at the estimated time-to-intercept.

The algorithm checks the following before outputting a valid launch plan:
1. The distance from the predicted target position to the intercept position must be less than the distance from the initial target position to the intercept position.  Otherwise, the target is flying away from the intercept position.
2. The distance from the intercept point, i.e., the output of the 2D interpolator, to the predicted position must be within some threshold.  Otherwise, the launch planner should not trust the interpolator that the interpolated value is accurate.


daniellovell (approved):
Approving since the merge catastrophe I created is now sorted out 👍

PR #23, opened 2025-01-10
Add tactical view (2D) panel with APP6-D symbology

daniellovell:
Involves some IADS system implementation that is used to render the UI improvements.

Allows users to hit TAB key to switch between the 3D view and the new "Tactical view" which implements APP-6(D) NATO Joint Military Symbology. The "tactical view" is designed to be somewhat representative of C2 (Command & control) battlespace monitoring systems that may be used in integrated air defense systems. 

Unity_zzT7lYTsHO

The IADS is modified to utilize a new TrackFileData data structure. TrackFileData is supposed to model the concept of [radar track files](https://www.radartutorial.eu/10.processing/sp21.en.html).

```
public abstract class TrackFileData {
  public Agent Agent { get; protected set; }
  [SerializeField]
  protected TrackStatus _status;
  public string TrackID { get; protected set; }

  public TrackStatus Status => _status;

  protected TrackFileData(Agent agent, string trackID) {
    Agent = agent;
    _status = TrackStatus.UNASSIGNED;
    TrackID = trackID;
  }

  public virtual void MarkDestroyed() {
    _status = TrackStatus.DESTROYED;
  }
}
```


Assets/Scripts/IADS/IADS.cs
```diff
@@ -23,6 +27,8 @@ public class IADS : MonoBehaviour {
       new Dictionary<Interceptor, Cluster>();
   private HashSet<Interceptor> _assignableInterceptors = new HashSet<Interceptor>();
 
+  private int _trackFileCount = 0;
```
tryuan99, line 30: Maybe use `_trackFiles.Count` instead?

Assets/Scripts/IADS/TrackFileData.cs
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public virtual void MarkDestroyed() {
+    _status = TrackStatus.DESTROYED;
+  }
+}
+
+[System.Serializable]
+public class ThreatData : TrackFileData {
+  public int assignedInterceptorCount = 0;
```
tryuan99, line 28: Return `_assignedInterceptors.Count` instead of maintaining its own state

Assets/Scripts/IADS/IADS.cs
```diff
@@ -15,6 +15,10 @@ public class IADS : MonoBehaviour {
   private IAssignment _assignmentScheme = new MaxSpeedAssignment();
 
   [SerializeField]
+  private List<TrackFileData> _trackFiles = new List<TrackFileData>();
+  private Dictionary<Agent, TrackFileData> _trackFileMap = new Dictionary<Agent, TrackFileData>();
+
+  private List<Interceptor> _assignmentQueue = new List<Interceptor>();
   private List<Threat> _threats = new List<Threat>();
   private List<Cluster> _threatClusters = new List<Cluster>();
```
tryuan99, line 23: Are these still used?  Looks like they're removed from all the methods below

PR #34, opened 2025-01-31
[IADS] Recluster missed and escaping threats

tryuan99:
This PR has not been extensively and tested and should be prior to merging.

This PR defines new coroutines in the IADS to execute the following rather computationally heavy functions:
1. Every 0.4 seconds, the IADS runs the planner and the predictor on the clusters to check whether it should launch a carrier interceptor towards a cluster.
2. Every 5 seconds, the IADS checks for escaping threats, which are threats that are currently assigned to some interceptors but that will likely out-run the pursuing interceptors.
3. Every 2 seconds, the IADS clusters threats that need to be reclustered.

Cluster reclustering can happen either because the threat was missed by the assigned interceptors and has no assigned interceptors anymore or because the threat managed to escape all of its assigned interceptors.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         testMode:
-          - playmode
+          # - playmode
           - editmode
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


daniellovell (approved):
Fantastic, fully autonomous sim!

PR #38, opened 2025-02-25
[Assignment] Add plugin for assignment solvers

tryuan99:
This PR changes the `MaxSpeedAssignment` to use the C++-based optimization tools offered by OR-Tools, the toolbox made by Google's operations research team (https://developers.google.com/optimization).  The cost-based assignment scheme is solved by formulating the problem as a constraint programming problem solved by CP-SAT (constraint programming using satisfiability methods) from OR-Tools.

The cover assignment guarantees that at least one interceptor is assigned to every threat.  The even assignment goes a step further and guarantees that the interceptors are evenly distributed among all threats.
The benefit of an even assignment scheme is visible when seven interceptors are assigned to two threats.  With a cover assignment scheme, usually six of the seven interceptors are assigned to one threat with one interceptor pursuing the lone "outlier" threat.

**Cover assignment**:
image
**Even assignment**:
image

The shared libraries are from the following commit: https://github.com/PisterLab/micromissiles-unity-plugins/commit/44388ae808e7bbf2437dc8a57353b4a695567a59.


daniellovell:
Failing some tests + lets integrate the plugins https://github.com/PisterLab/micromissiles-unity-plugins/ directly into the monorepo

tryuan99:
Fixed the unit tests:
- The interceptors in the `MaxSpeedAssignmentTest` did not have an acceleration configuration, so the tests would error out when accessing the parameters of the interceptors' static acceleration configurations.
- `CalculateMaxForwardAcceleration` and `CalculateMaxNormalAcceleration` are now public methods, so they should not be invoked with `InvokePrivateMethod` in the `ThreatTests`.

I agree with making a monorepo, but I would like to do that in a separate PR.  In any case, the shared libraries need to be copied out of the `bazel-bin` into a path that is accessible by Unity.

tryuan99:
@daniellovell Could you help me verify whether the .dll plugin generated by GitHub Action works for you please?

tryuan99:
Looks like the plugin should be compiled on Ubuntu 22.04 to maintain compatibility with the `glibc` and `libstdc++` standard libraries.  Previously, running Unity on Ubuntu 22.04 would throw a `DllImportException` and `ldd` produces an error:
```bash
$ ldd Assets/Plugins/libassignment.so
Assets/Plugins/libassignment.so: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by Assets/Plugins/libassignment.so)
Assets/Plugins/libassignment.so: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by Assets/Plugins/libassignment.so)
Assets/Plugins/libassignment.so: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.31' not found (required by Assets/Plugins/libassignment.so)
	linux-vdso.so.1 (0x00007ffe9dc6b000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000071410f119000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x000071410ee00000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000071410fdf7000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000071410ea00000)
	/lib64/ld-linux-x86-64.so.2 (0x000071410fe2d000)

```
After compiling the plugin on Ubuntu 22.04, `ldd` seems happy:
```bash
$ ldd Assets/Plugins/libassignment.so
	linux-vdso.so.1 (0x00007fff15526000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000070697144f000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x0000706970400000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000070697142f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000706970000000)
	/lib64/ld-linux-x86-64.so.2 (0x000070697154c000)
```

PR #53, opened 2025-07-22
[Actions] Deploy docs on push to master only

tryuan99:
This PR fixes the loophole where any bad actor could create a PR with incorrect documentation changes, and it'll be public immediately.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -16,10 +16,6 @@ permissions:
 
 jobs:
   build:
-    permissions:
-      contents: write
-      pages: write
-      id-token: write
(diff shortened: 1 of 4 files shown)


.github/workflows/docs.yaml
```diff
@@ -7,19 +7,16 @@ on:
   pull_request:
     branches:
       - master
-  workflow_dispatch:  # Allows manual triggering.
```
daniellovell, line 10: Why remove comment? could be useful for reference

daniellovell (commented):
lgtm, thanks!

daniellovell (approved):
Ship it 🫡

PR #54, opened 2025-07-23
[Docs] Fix deploying documentation

tryuan99:
My previous PR #53 broke the documentation deployment job because I separated the build and deploy jobs.  This PR fixes that and also cleans up the action.  See https://github.com/peaceiris/actions-gh-pages for more details on the action options.
* The GitHub token already has the following permissions, so we don't need to specify more:
  ```
  GITHUB_TOKEN Permissions
  Contents: write
  Metadata: read
  Pages: write
  ```
* `allow_empty_commit` is not needed to prevent spamming the `gh-pages` branch.
* `keep_files` should be removed because older files will not be removed.  For example, `Cpp_Plugins_Guide.html` still exists after renaming the page to `Plugins_Guide.html`.
* `force_orphan` is false by default.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -9,15 +9,12 @@ on:
       - master
   workflow_dispatch:  # Allow manual triggering.
 
-permissions:
-  contents: write
-  pages: write
-  id-token: write
-
 jobs:
-  build:
-    name: Build documentation
+  build-deploy:
+    name: Build and deploy documentation
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yaml
```diff
@@ -33,21 +30,9 @@ jobs:
...
-    runs-on: ubuntu-latest
-    steps:
       - name: Deploy to GitHub Pages
         uses: peaceiris/actions-gh-pages@v4
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs/.vitepress/dist
-          allow_empty_commit: true
-          keep_files: true
```
daniellovell, line 52: We should add keep_files back in, and a comment explaining that it's necessary for the coverage reports generated by the testing Action to be preserved by the documentation deployment Action

tryuan99: That smells a bit fishy to me.  I attempted to circumvent this problem by having the documentation workflow run after the test workflow has completed (since the coverage reports are needed as part of the documentation).

tryuan99:
@daniellovell I can't figure out why the `Documentation` workflow isn't running on this PR even though I set it to run after the `Test` workflow has completed.

daniellovell:
I believe I have fixed the issue you encountered @tryuan99 because the documentation builds and did NOT deploy in the context of this PR. But we should merge the PR to check whether deployment is properly triggered after the merge

image

PR #55, opened 2025-08-05
[Config] Add Protobuf configurations

tryuan99:
This PR introduces Protobuf configurations to supersede the current JSON configurations.  The configurations are specified in Protobuf text format (`.pbtxt`), which are parsed according to the protos.  The two main configuration protos are `StaticConfig`, used to define the properties of each agent, and the `SimulatorConfig`, used to define the engagement scenario.

Note that the Protobuf plugin must be compiled with `bazel build -c opt //:protobuf` to prevent a memory alignment issue, detailed below.

All `.cs` files in this PR are generated files and thus also excluded from the linter.  The `.proto` files define the Protobuf message structure and are shared by both Unity C# and C++.

Example usage code:
```csharp
    int size = 100000;
    byte[] buffer = new byte[size];
    int serializedLength = 0;
    fixed (void* bufferPtr = buffer)
    {
      serializedLength = Protobuf.Protobuf_StaticConfig_LoadToBinary("configs/interceptors/micromissile.pbtxt", (IntPtr)bufferPtr, size);
    }
    Debug.Log($"serialized length {serializedLength}");
    Micromissiles.StaticConfig cs = Micromissiles.StaticConfig.Parser.ParseFrom(buffer, 0, serializedLength);
    Debug.Log($"{cs}");
```

Log output:
image


tryuan99:
Looks like an optimized build does work though.  Using the plugin libraries built through GitHub Actions, the Protobuf plugin successfully parses the Protobuf text file and sends it to the simulator:
image

configs/interceptors/hydra70.pbtxt
daniellovell, line 1: Files located in the root `configs/...` folder won't be included in builds of the Unity Player (the shippable application). These should be located in `Assets/StreamingAssets/...` if we want them to be built into the shipped app.

Assets/Scripts/Generated.meta
daniellovell, line 1: All the C# generated files are now in this `Asset/Scripts/Generated/` directory. Why don't we encapsulate these C# files into its own dedicated `.asmdef` that references `Google.Protobuf` That way warnings are isolated. Also it should help compile time and it's just better dependency hygiene.

configs/interceptors/hydra70.pbtxt
tryuan99: I'm addressing this in a future PR.

PR #56, opened 2025-08-06
[Config] Load simulator config from Protobuf text format

tryuan99:
This PR demonstrates how the simulator configuration is defined in a Protobuf text file `.pbtxt` and loaded through the Protobuf plugin into the Unity-based simulator.  This PR was tested by simply running the simulation and ensuring that nothing crashes.


daniellovell (dismissed):
edit

PR #57, opened 2025-08-06
[Actions] Trigger build on PR to master but not release

daniellovell:
Add master branch to Build workflow triggers
Restrict Release workflow to only trigger on version tags

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,11 +4,13 @@ on:
   push:
     branches:
       - release
+      - master
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+'
   pull_request:
     branches:
       - release
+      - master
   schedule:
     - cron: '0 2 * * *'  # Run at 2 AM UTC every day.
   workflow_dispatch:  # Allow manual triggering.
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -4,11 +4,13 @@ on:
   push:
     branches:
       - release
+      - master
```
tryuan99, line 7: super super minor nit: order the branch names alphabetically for easier maintainability

.github/workflows/release.yaml
```diff
@@ -4,16 +4,15 @@ on:
...
 
 jobs:
   release:
     name: Create release
     runs-on: ubuntu-latest
     if: |
       github.event.workflow_run.conclusion == 'success' &&
-      github.event.workflow_run.event != 'schedule'
+      github.event.workflow_run.event == 'push' &&
+      startsWith(github.event.workflow_run.head_branch, 'v')
```
tryuan99, line 15: perhaps:
```
github.event.workflow_run.head_branch matches '^v\d+(\.\d+)*$'
```
?

daniellovell: Supposedly regex not allowed (sources: https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#startswith https://github.com/orgs/community/discussions/26186) , but you drew my attention to the fact that it needs to be `refs/tags/v` instead of `v`

PR #60, opened 2025-09-04
[Config] Migrate static configs to Protobuf text format

tryuan99:
This PR migrates all static configurations for interceptors and threats to Protobuf.  The static configuration proto definition has been updated, the static configurations are now in Protobuf text format, and the `ConfigLoader` now loads the Protobuf text format directly.  The `plugins` Bazel repository declares a local repository dependency on the `proto` definitions in `Assets/Proto` and on the `pbtxt` files in `Assets/StreamingAssets/Configs`.

Unusually, the Protobuf C# code is different from the C++ code in that any unpopulated fields return `null` on accessing them (whereas in C++, unpopulated fields return their default value when accessing them).  This is why the static `ProtobufInitializer` class serves to populate all unpopulated fields with the default "zero" value.

This PR is big, but a lot of it is generated code (generated Protobuf C# code) or configuration changes.


plugins/.gitignore
```diff
@@ -1,8 +1,2 @@
-# Mac.
-**/.DS_Store
-
 # Bazel.
 /bazel-*
-
-# VS Code.
-.vscode
```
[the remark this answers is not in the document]
tryuan99: Already ignored by global `.gitignore`

Assets/StreamingAssets/Configs/Models/ascm.pbtxt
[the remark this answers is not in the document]
tryuan99: When others add their own configs, they'll have to prepend `quadcopter.pbtxt` with `Threats/quadcopter.pbtxt` because all static configurations are loaded by the same function.

I won't be able to tell in software whether the newly added static config is an interceptor or threat, so I decided in favor of better UX to not differentiate interceptors from threats.

Assets/Scripts/Behavior/AttackBehavior.cs
```diff
@@ -13,9 +13,9 @@ public class AttackBehavior {
   public FlightPlan flightPlan;
   // Returns the next waypoint for the threat to navigate to
   // In addition, return the power setting to use toward the waypoint
-  public virtual (Vector3 waypointPosition, PowerSetting power)
+  public virtual (Vector3 waypointPosition, Micromissiles.Power power)
```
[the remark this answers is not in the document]
tryuan99: Restructured `Assets/Proto` directory - namespaces now correspond to what the proto messages are used for like `Configs` or `Simulation` (for coordinates and other simulation utilities).

tryuan99:
Build and test both pass: https://github.com/PisterLab/micromissiles-unity/actions/runs/17534572673, https://github.com/PisterLab/micromissiles-unity/actions/runs/17534574011

Assets/Scripts/Config/ConfigLoader.cs
```diff
@@ -12,6 +13,22 @@ public static class ConfigLoader {
...
+
+  // Map from the threat type to the static configuration file.
+  private static readonly Dictionary<Configs.ThreatType, string> ThreatStaticConfigMap = new() {
+    { Configs.ThreatType.Quadcopter, "quadcopter.pbtxt" },
+    { Configs.ThreatType.Ucav, "ucav.pbtxt" },
+    { Configs.ThreatType.Brahmos, "brahmos.pbtxt" },
+    { Configs.ThreatType.Ascm, "ascm.pbtxt" },
+    { Configs.ThreatType.Fateh110B, "fateh_110b.pbtxt" },
+  };
+
```
daniellovell, line 31: Hmm this is showing compile-time binding of static configurations right? I think this is counter to our philosophy of users being able to add their own models when they are working with just the binary.

Like Kris will not be able to make and add his own model files anymore after this change?

Assets/Proto/Configs/agent.proto
```diff
@@ -1,20 +1,25 @@
...
 enum ThreatType {
-  QUADCOPTER = 0;
-  UCAV = 1;
+  INVALID_THREAT = 0;
+  QUADCOPTER = 1;
+  UCAV = 2;
+  BRAHMOS = 3;
+  ASCM = 4;
+  FATEH_110B = 5;
 }
```
daniellovell, line 23: This enum is user-extendable after compilation, right? But I'm concerned that the Unity-side implementation will not support us adding additional Threats and Interceptors to the agent.proto enum.

daniellovell (changes requested):
Tested and working well on Windows :) 

Mainly concerned about the compile-time binding of static configurations between types/files in the ConfigLoader. I think this is against our core design philosophy.

Assets/Proto/Configs/agent.proto
tryuan99: Not extendable because the Protobuf is compiled.

Assets/Scripts/Config/ConfigLoader.cs
tryuan99: Let me address this in the simulation config PR - the static config PR does not implement compile-time binding of static configs.

daniellovell: Got it, that makes sense to me 👍

Assets/Proto/Configs/agent.proto
daniellovell: Will the `oneof` construct allow users to add more threats at runtime from the files present? 
Also seems kinda painful to maintain cause every time we add a new threat or interceptor model gotta go update the enum..

tryuan99: Will remove all enums and just refer to static configurations by filename.

tryuan99:
Build: https://github.com/PisterLab/micromissiles-unity/actions/runs/17870255612
Test: https://github.com/PisterLab/micromissiles-unity/actions/runs/17870258524

PR #63, opened 2025-09-23
[Config] Handle Protobuf plugin errors

tryuan99:
Previously, the Protobuf plugin was implemented to throw exceptions for all kinds of errors, including invalid filenames, insufficient buffer size, and syntax errors in the Protobuf text files.  This caused the Unity to crash whenever the plugin threw an exception.

This PR adds a `StatusCode` enum and `Status` proto message.  These are intended for a clean status interface between the plugins and Unity and prevent Unity from crashing when a plugin error occurs.  This PR also modifies the Unity code to check for the status code after executing a plugin function call.  This is not the optimal solution because:
  1. Unity only knows the status code of any plugin function call
  2. any error messages are printed to `stderr` in the plugin, which Unity does not capture.
However, this is a much cleaner solution than throwing exceptions in the plugin.

In the future, the status codes should be used by the assignment plugin, too, in case unexpected errors occur during execution.

Build: https://github.com/PisterLab/micromissiles-unity/actions/runs/17960155816
Test: https://github.com/PisterLab/micromissiles-unity/actions/runs/17960158769


daniellovell (approved):
Tested and seems to be working well for me. Changes look good.

PR #65, opened 2025-09-24
[Configs] Add Protobuf plugin function to get serialized message length

tryuan99:
As title.  This was suggested in https://github.com/PisterLab/micromissiles-unity/pull/55#pullrequestreview-3098522792, and I think this is a cleaner solution than dynamically increasing the serialized buffer size until the entire Protobuf message fits.


tryuan99:
Build: https://github.com/PisterLab/micromissiles-unity/actions/runs/17991459426
Test: https://github.com/PisterLab/micromissiles-unity/actions/runs/17991462080

Assets/Scripts/Config/ConfigLoader.cs
```diff
@@ -38,46 +34,52 @@ public static string LoadFromStreamingAssets(string relativePath) {
...
+      int* serializedLengthPtr = &serializedLength;
+      fixed(void* bufferPtr = buffer) {
+        status = loadFunction(streamingAssetsPath, (IntPtr)bufferPtr, serializedLength,
+                              (IntPtr)serializedLengthPtr);
       }
+    }
+    if (status != Plugin.StatusCode.StatusOk) {
       Debug.Log($"Failed to load the Protobuf text file {relativePath} with status code {status}.");
       return default;
     }
```
daniellovell, line 67: Kind of a minor issue, but the new flow exits immediately whenever `Protobuf_*_LoadToBinary` reports `StatusFailedPrecondition`. That status still happens if the pbtxt changes between the length probe and the load (e.g., someone Ctrl+S's some changes to the file while Unity is still loading the config). 

Before this PR we would retry with a larger buffer and eventually succeed. now we just log and return null, regressing the file-saving-while-running scenario.

Might consider restoring the retry logic that used to be there (i.e. re-run the size-query and retry the load once more; if fail twice, give up)

daniellovell (approved):
One minor comment below, probably optional to address so marking approve cause otherwise looks good & tests good.

Issue #69, opened 2025-10-03
No gravity on Hydra 70 interceptors


tryuan99:
Investigate why there's no gravity acting on the Hydra 70 interceptors.

<img width="300" height="334" alt="Image" src="https://github.com/user-attachments/assets/b211eae7-faca-4ca8-9b0f-e8ca4e1d68b3" />

<img width="667" height="308" alt="Image" src="https://github.com/user-attachments/assets/fcf67cb1-f833-4f08-a74e-ebeddac104f9" />

The acceleration should be at least 9.81 in the y-direction.

PR #70, opened 2025-10-04
[Bugfix] Ensure idle interceptors experience gravity & avoid null ref excepts in target model update

daniellovell:
Idle controller input: When an interceptor has no assigned target, we now return zero control input. Gravity and drag are still applied centrally in `AerialAgent.CalculateAcceleration`, so missiles/carriers fall naturally instead of hovering.

UpdateTargetModel now returns early when target/model/sensor config is missing to prevent null reference exceptions during phases before assignment or with incomplete configs.

After fix:
image

Addresses #69  (lol)


tryuan99 (approved):
Tested on my end and works as expected.  Thanks!

PR #76, opened 2025-10-06
[Bazel/Pre-Commit] Bump Bazel dependency and pre-commit hook versions

tryuan99:
As title.  This PR contains the following changes:
 1. The Bazel version has been bumped to 8.4.2, which adds back support for MSVC + Bazel.
 2. The Bazel dependency versions have been bumped.
 3. The pre-commit hook versions have been bumped.
 4. All pre-commit hooks were run on all existing files, so some JSON files got reformatted.


tryuan99:
The `protobuf_test` failure is unrelated to the Bazel version bump and is already present on the `master` branch.

tryuan99:
The protobuf_test failure is unrelated to the Bazel version bump and is already present on the master branch.

@daniellovell I would like to merge all these PRs into `master` (along with your addition of running the `plugins` Bazel tests on Windows) and then fix the Bazel test regression.  My guess is that #78 changes the Protobuf API to use `std::filesystem::path`, and that should fix the path handling on Windows.

PR #77, opened 2025-10-06
[Assignment] Return status codes from assignment plugin

tryuan99:
This PR fixes the assignment plugin to return status codes, just like the Protobuf plugin, instead of throwing exceptions on the C++ side, which causes the simulation to crash.  This PR also adds unit tests for invalid costs and weights to test the failed status codes.


daniellovell (approved):
Implementation LGTM, two higher level requests though - 
For documentation could we add to the Plugins_Guide.md examples/explanation for the error handling patterns for plugins, how to check status codes from Unity, and the common status codes and their meanings?

For test coverage, the tests in `cover_assignment_test.cc` only test valid inputs.

I think a Unity editor test that exercises the new error handling pattern would be useful too. I.e. something that passes invalid data to assignment plugin -> verifies simulation doesn't crash -> confirms error is logged

PR #83, opened 2025-10-07
[Plugins] Remove unsafe blocks from plugin function calls

### LINKED ISSUE #82: [Plugins] Remove unsafe block
Plugins currently use `IntPtr` in an `unsafe` block to pass output argument arrays to Unity.

See https://github.com/PisterLab/micromissiles-unity/pull/81#discussion_r2409222059 and https://github.com/PisterLab/micromissiles-unity/pull/81#discussion_r2409222061 for suggestions on how to remove the `unsafe` block.

tryuan99:
Previously, we used `IntPtr` to pass memory buffers between Unity and the plugins, which necessitated an `unsafe` block.  However, this is not unnecessary, and it is much more concise and safer to simply pass arrays to the plugins and use `out int` for output integer arguments.  This PR resolves #82.


daniellovell (approved):
Looks good - can't even find anything to nitpick.

Issue #85, opened 2025-10-07
[Docs] Coverage reports are returning 404


tryuan99:
<img width="1420" height="531" alt="Image" src="https://github.com/user-attachments/assets/0e44a4d2-b03f-4e80-af64-d17e66d7154a" />

tryuan99:
Bisecting gives this commit as the culprit: https://github.com/PisterLab/micromissiles-unity/commit/27adf6a50345e3b095cd31dcfc3d293bff0e3e41

daniellovell:
Re-opening issue, still not fixed by #86

PR #90, opened 2025-10-08
[Docs] Download coverage reports into individual directories

tryuan99:
As title.  This PR changes the artifact names of the uploaded coverage reports to `coverage-editmode` and `coverage-playmode`.  When downloading the artifacts to build the documentation, the coverage reports will be downloaded to individual directories, `coverage/coverage-editmode` and `coverage/coverage-playmode`.


tryuan99:
Merging to see if this resolves #85.

PR #96, opened 2025-10-12
[Interceptor/Threat] Add interceptor and threat interfaces

tryuan99:
This PR adds the interceptor and threat interfaces and base implementations as well as some of the related classes.  Again, no effect on the actual simulator, tested by running unit tests.


daniellovell (approved):
Approved 👍

PR #97, opened 2025-10-12
[Algorithm] Add algorithm interfaces and base implementations

tryuan99:
This PR adds the swarm algorithm interfaces and base implementations, according to the design doc (https://docs.google.com/document/d/1FINN_yEZ02AkopUn-ev0bzoi4oKenHYyIpEO6bSFrlE/edit?usp=sharing).  There is no effect on the current state of the simulator as I have just renamed existing conflicting classnames and struct names with a `Legacy` suffix.  The only unit-testable class is the `Cluster` class, so I have added unit tests for that class.


Assets/Scripts/Algorithms/Prediction/PredictorState.cs
```diff
@@ -0,0 +1,13 @@
...
+
+// The predictor state defines the output of a predictor.
+public class PredictorState {
+  // Position.
+  public Vector3 Position { get; set; }
+
+  // Velocity.
+  public Vector3 Velocity { get; set; }
+
+  // Acceleration.
```
daniellovell: nitpick but the periods after one word look awkward to me here

daniellovell (approved):

PR #98, opened 2025-10-13
[Agent] Fix lift acceleration calculation

tryuan99:
The lift acceleration is any acceleration that does not point in the direction of the agent's velocity vector.  Calculating the lift-induced drag has now been corrected to project the acceleration input onto the plane whose normal vector points along the velocity vector.


daniellovell (approved):
I agree. As a sanity check, in the screenshot below, `transform.forward` is blue, `transform.right` is red, and `transform.up` is green. 

Image

PR #100, opened 2025-10-13
[Sensor] Add sensor interface and ideal sensor implementation

tryuan99:
This PR adds the `ISensor` and `SensorBase` implementations and changes the sensors to not inherit from `MonoBehaviour`.  The old sensor classes have been renamed with a `Legacy` suffix and are slated for deletion after the complete hierarchy implementation.

This PR also changes the sensor output (and transformation) structs to adhere to the style guide of having capitalized public attributes.

Unit tests for the ideal sensor are provided although since the implementation uses the agent's `GetRelativeTransformation`, there is not much to test.


Assets/Scripts/Managers/SimManager.cs
```diff
@@ -394,7 +394,7 @@ public Interceptor CreateInterceptor(Configs.AgentConfig config, Simulation.Stat
     // Create a sensor for the interceptor.
     switch (config.DynamicConfig.SensorConfig.Type) {
       case Simulation.SensorType.Ideal: {
-        interceptorObject.AddComponent<IdealSensor>();
+        interceptorObject.AddComponent<IdealSensorLegacy>();
         break;
       }
       default: {
```
daniellovell: Do we want to consider a MonoBehaviour wrapper? e.g.
```cs
public class IdealSensorComponent : MonoBehaviour {
    private IdealSensor _sensor;

    private void Awake() {
      var agent = GetComponent<IAgent>();
      _sensor = new IdealSensor(agent);
    }

    public SensorOutput Sense(IHierarchical hierarchical) => _sensor.Sense(hierarchical);
    public SensorOutput Sense(in Vector3 waypoint) => _sensor.Sense(waypoint);
}
```

Or shall we store sensors as fields?
```cs
// In Interceptor or Agent class
private ISensor _sensor;

void Initialize() {
_sensor = new IdealSensor(this);
}
```

Assets/Tests/EditMode/IdealSensorTests.cs
daniellovell: We're only looking at basic correctness here. Worth considering some edge cases.

Assets/Scripts/Sensors/IdealSensor.cs
```diff
@@ -1,25 +1,28 @@
 using UnityEngine;
 
-public class IdealSensor : Sensor {
-  public override SensorOutput Sense(Agent agent) {
-    SensorOutput agentSensorOutput = new SensorOutput();
+// Ideal sensor.
+//
+// The ideal sensor does not add any noise to the measurements.
```
daniellovell:
```suggestion
// Ideal sensor implementation.
//
// The ideal sensor provides perfect, noise-free measurements of the relative transformation
// between the sensing agent and a target (either an IHierarchical object or a waypoint).
// This sensor is useful for testing and establishing baseline behavior before adding
// realistic sensor noise and limitations
```

Assets/Tests/EditMode/IdealSensorTests.cs
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  [Test]
+  public void Sense_WaypointReturnsSameAsRelativeTransformation() {
+    var waypoint = new Vector3(-12, 20, -1);
+    var relativeTransformation = _agent.GetRelativeTransformation(waypoint);
+    var sensorOutput = _sensor.Sense(waypoint);
+    Assert.AreEqual(relativeTransformation.Position, sensorOutput.Position);
+    Assert.AreEqual(relativeTransformation.Velocity, sensorOutput.Velocity);
+  }
```
daniellovell: The function names for the tests here describe the implementation rather than the behavior that's tested. The convention for naming tests is the behavior that's being tested so the function names double as a specification.

Assets/Scripts/Sensors/SensorOutput.cs
```diff
@@ -0,0 +1,9 @@
+// The sensor output defines the output of a sensor. The transformations are relative to the sensing
+// agent's state.
+public struct SensorOutput {
+  // Relative position transformation.
+  public PositionTransformation Position { get; set; }
+
+  // Relative velocity transformation.
+  public VelocityTransformation Velocity { get; set; }
```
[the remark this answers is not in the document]
daniellovell: Suggest:
```cs
public struct SensorOutput {
    // Relative position transformation.
    public PositionTransformation Position { get; init; }

    // Relative velocity transformation.
    public VelocityTransformation Velocity { get; init; }
}
```

Which would allow
```cs
// Allowed
var output = new SensorOutput { Position = pos, Velocity = vel };

// Compilation error
output.Position = newPos;  // Error: init-only property
```

Assets/Scripts/Managers/SimManager.cs
tryuan99: I'm leaning towards sensors as fields; I don't see a benefit to a wrapper, and it makes unit-testing them harder.

daniellovell (approved):
Resolution looks good and #116 covers the addition of `init` accessors. lgtm

PR #102, opened 2025-10-14
[Evasion] Implement evasion behaviors

tryuan99:
This PR implements the evasion interface as well as the `NoEvasion` and the `OrthogonalEvasion`.  The `Evasion` interface has been slightly modified to store the agent as a private variable.  Unit tests are provided as well.

Note before merging: plugins have to be regenerated because I modified the `dynamic_config.proto` definition.


Assets/Scripts/Evasion/OrthogonalEvasion.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    // Evade the pursuer by turning the velocity to be orthogonal to the pursuer's velocity.
+    Vector3 normalVelocity = Vector3.ProjectOnPlane(agentVelocity, pursuerVelocity);
+    if (normalVelocity.sqrMagnitude < epsilon) {
+      // If the agent's velocity is aligned with the pursuer's velocity, choose a random normal
+      // direction in which to evade.
+      normalVelocity = pursuer.transform.right;
+    }
+    Vector3 normalAccelerationDirection =
+        Vector3.ProjectOnPlane(normalVelocity, agentVelocity).normalized;
```
daniellovell: What happens when the Interceptor's right vector is aligned with the Threat's velocity? Wont this projection cause a Vector3.zero?

Should we guard this?

```cs
if (normalVelocity.sqrMagnitude < epsilon) {
    // Choose a direction perpendicular to agent velocity
    normalVelocity = Vector3.Cross(agentVelocity, Vector3.up);
    if (normalVelocity.sqrMagnitude < epsilon) {
        // If velocity is vertical, use forward as reference
        normalVelocity = Vector3.Cross(agentVelocity, Vector3.forward);
    }
}
Vector3 normalAccelerationDirection = Vector3.ProjectOnPlane(normalVelocity, agentVelocity);
if (normalAccelerationDirection.sqrMagnitude < epsilon) {
    Debug.LogWarning("Failed to compute evasion direction");
    return Vector3.zero;
}
normalAccelerationDirection = normalAccelerationDirection.normalized;
```

Assets/Scripts/Agents/AgentBase.cs
```diff
@@ -76,6 +79,17 @@ private void Awake() {
...
 
+  public float MaxForwardAcceleration() {
+    return StaticConfig.AccelerationConfig?.MaxForwardAcceleration ?? 0;
+  }
+
+  public float MaxNormalAcceleration() {
+    float maxReferenceNormalAcceleration =
+        (StaticConfig.AccelerationConfig?.MaxReferenceNormalAcceleration ?? 0) * Constants.kGravity;
+    float referenceSpeed = StaticConfig.AccelerationConfig?.ReferenceSpeed ?? 1;
+    return Mathf.Pow(Speed / referenceSpeed, 2) * maxReferenceNormalAcceleration;
```
daniellovell: Division by zero if `referenceSpeed` is zero.

Assets/Scripts/Evasion/OrthogonalEvasion.cs
```diff
@@ -0,0 +1,86 @@
...
+  public override bool ShouldEvade(IAgent pursuer) {
+    // Default evasion range threshold in meters.
+    const float defaultEvasionRangeThreshold = 1000f;
+
+    if (!(Agent.AgentConfig?.DynamicConfig?.FlightConfig?.EvasionConfig?.Enabled ?? false)) {
+      return false;
+    }
+
+    float evasionRangeThreshold =
+        Agent.AgentConfig?.DynamicConfig?.FlightConfig?.EvasionConfig?.RangeThreshold ??
```
daniellovell: On line 16 we already guarded the null EvasionConfig, so there is no need for this redundant null check.

Assets/Tests/EditMode/OrthogonalEvasionTests.cs
daniellovell: Recommend adding

```cs
[Test]
public void Evade_AgentAndPursuerAlignedVelocities_NoNaN() {
    _agent.Position = Vector3.zero;
    _agent.Velocity = new Vector3(0, 100, 0);
    _pursuer.Position = new Vector3(0, 100, 0);
    _pursuer.Velocity = new Vector3(0, 100, 0); // Aligned

    Vector3 result = _evasion.Evade(_pursuer);
    Assert.IsFalse(float.IsNaN(result.x));
    Assert.IsFalse(float.IsNaN(result.y));
    Assert.IsFalse(float.IsNaN(result.z));
}
```

Assets/Proto/Configs/dynamic_config.proto
```diff
@@ -10,17 +10,17 @@ enum ControllerType {
...
+// Evasion configuration.
+message EvasionConfig {
+  // If true, evasion is enabled.
+  bool enabled = 1;
 
-    // Range threshold in m at which the agent should evade.
-    float range_threshold = 2;
-  }
+  // Range threshold in m at which the agent should evade.
+  float range_threshold = 2;
```
daniellovell: nit: recommend a little more clear documentation here
```cs
  // Range threshold in meters. The agent will evade when a pursuer is within this
  // range AND closing (velocity.Range < 0).
```

Assets/Tests/EditMode/OrthogonalEvasionTests.cs
daniellovell: I think we also would like a more integration-like test that verifies the forward acceleration calculation works with the power table
```cs
  [Test]
  public void Evade_ThreatAgent_IncludesForwardAcceleration() {
      // Create a mock threat agent with power table
      // Verify that the returned acceleration includes both normal and forward components
  }
```

Assets/Scripts/Evasion/OrthogonalEvasion.cs
tryuan99: If `normalAccelerationDirection`'s magnitude is too small, it means that the threat is orthogonal to the pursuer already.  In this case, it should not accelerate anywhere normally, and Vector3.zero is indeed the correct normal acceleration.

Assets/Scripts/Agents/AgentBase.cs
tryuan99: I'd argue that an exception should be thrown here because at a reference speed of 0, no normal acceleration should be possible.

Assets/Scripts/Evasion/OrthogonalEvasion.cs
```diff
@@ -0,0 +1,86 @@
...
+        defaultEvasionRangeThreshold;
+    SensorOutput sensorOutput = Agent.Sensor.Sense(pursuer);
+    return sensorOutput.Position.Range <= evasionRangeThreshold && sensorOutput.Velocity.Range < 0;
+  }
+
+  // Calculate the acceleration input to evade the pursuer.
+  public override Vector3 Evade(IAgent pursuer) {
+    const float epsilon = 1e-12f;
```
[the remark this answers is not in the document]
tryuan99: Sure, but I'm comparing Sqr against this value.

Assets/Scripts/Threats/ThreatBase.cs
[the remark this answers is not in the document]
tryuan99: Yes, this happens in later PR.

PR #108, opened 2025-10-17
[Predictor/Planner] Clean up predictor and planner classes

tryuan99:
This PR cleans up all predictor and planner interfaces and implementations to interface with `IHierarchical` and `IAgent`.  In particular:
* I cleaned up all the relevant `Utils/*.cs` classes, including the `KDTree`, the `IInterpolator2D`, and `NearestNeighborInterpolator2D`.
* I cleaned up predictor classes, including the linear extrapolator.
* I cleaned up all planner classes, including all the subclasses of `ILaunchPlanner` (`IterativeLaunchPlanner`) as well as of `ILaunchAnglePlanner`, including `LaunchAngleInterpolatorBase`, `LaunchAngleCsvInterpolator`, and `LaunchAngleDataInterpolator`.
* I added unit tests for all classes/revamped all of them to be more comprehensive, e.g., for the `InterceptPosition` function in the `LaunchAnglePlanner`.

All `*Legacy.cs` files are simply the existing files with the `Legacy` suffix to not break the current simulator.


Assets/Scripts/Algorithms/Planning/LaunchAngleInterpolator.cs
```diff
@@ -4,14 +4,32 @@
...
+  }
+
+  // Get the intercept position.
+  public Vector3 GetInterceptPosition(Vector3 position);
+
+  // Convert from a 3D vector to a 2D direction that ignores the azimuth.
+  protected static Vector2 ConvertToDirection(Vector3 position) {
+    return new Vector2(Vector3.ProjectOnPlane(position, Vector3.up).magnitude,
+                       Vector3.Project(position, Vector3.up).magnitude);
+  }
```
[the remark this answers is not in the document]
tryuan99: Comment is incorrect.  I also would like to keep it explicit here instead of using `position.y`.

Assets/Scripts/Utils/PriorityQueue.cs
```diff
@@ -26,11 +26,11 @@ public T Dequeue() {
...
 
-    float minKey = _buffer.Keys.Min();
-    Queue<T> queue = _buffer[minKey];
-    T item = queue.Dequeue();
+    var firstPair = _buffer.First();
+    var queue = firstPair.Value;
+    var item = queue.Dequeue();
     if (queue.Count == 0) {
-      _buffer.Remove(minKey);
+      _buffer.Remove(firstPair.Key);
```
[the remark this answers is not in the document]
tryuan99: No actionable item

daniellovell:
@coderabbit review 🥺

Assets/Scripts/Algorithms/Planning/IterativeLaunchPlanner.cs
```diff
@@ -66,14 +67,18 @@ public override LaunchPlan Plan() {
     Vector3 interceptorToInterceptPosition = interceptPosition;
     Vector3 threatToPredictedPosition = targetPosition - initialState.Position;
     if (Vector3.Dot(interceptorToInterceptPosition, threatToPredictedPosition) > 0) {
-      return LaunchPlan.NoLaunch;
+      return LaunchPlan.NoLaunch();
     }
```
daniellovell: `LaunchAnglePlanner.InterceptPosition` now returns an absolute world-space point offset by `Agent.Position`. I think using that absolute coordinate directly makes this dot product depend on the global origin instead of the launcher. 
e.g. place the launcher at (1000,0,0) and an intercept 50 m ahead, so the true relative vector is (50,0,0). The code here uses (1050,0,0).

This fix I believe is to derive the relative vector from the launcher `Vector3 interceptorToInterceptPosition = interceptPosition - LaunchAnglePlanner.Agent.Position`

Assets/Tests/EditMode/NearestNeighborInterpolator2DTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using UnityEngine;
+using UnityEngine.TestTools;
+using UnityEngine.TestTools.Utils;
+
+public class NearestNeighborInterpolator2DTests {
+  private const float _epsilon = 1e-3f;
+
+  [Test]
+  public void Interpolate_InvalidPoints_ReturnsEmptyResult() {
+    LogAssert.ignoreFailingMessages = true;
```
daniellovell: I believe this is a global static flag, so once a test sets it to true, all other tests afterwards will silently swallow their errors.

[LogAssert.Expect](https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/api/UnityEngine.TestTools.LogAssert.html#UnityEngine_TestTools_LogAssert_Expect_UnityEngine_LogType_System_String_) would be better here.

Assets/Scripts/Algorithms/Planning/LaunchAnglePlannerBase.cs
```diff
@@ -0,0 +1,30 @@
...
+
+  // Convert from a 3D vector to a 2D direction that ignores the azimuth and that is relative to the
+  // agent's position.
+  protected Direction ConvertToRelativeDirection(in Vector3 position) {
+    Vector3 relativePosition = position - Agent.Position;
+    return new Direction { Distance =
+                               Vector3.ProjectOnPlane(relativePosition, Vector3.up).magnitude,
+                           Altitude = Vector3.Project(relativePosition, Vector3.up).magnitude };
+  }
+}
```
daniellovell: Are we sure we don't want to allow launch angle planning for targets **below** the launcher? The `.magnitude` here will not allow that - sign information is lost
Recommend instead 
```cs
Altitude = Vector3.Project(relativePosition, Vector3.up).y  // Use .y instead of .magnitude
```

Assets/Scripts/Algorithms/Planning/LaunchPlan.cs
daniellovell: LaunchPlan is a struct with properties that have setters. When creating structs with object initializers, if any property setter has a bug or if the struct isn't properly initialized, you might get default values instead of the assigned values. also - struct fields not explicitly set will be zero-initialized.

This is generally why structs with mutable properties are an anti-pattern in C#, see SO thread below:

https://stackoverflow.com/questions/441309/why-are-mutable-structs-evil

I recommend either making it a class or making it immutable struct with constructor that forces all fields to be initialized.

Assets/Tests/EditMode/LaunchAngleDataInterpolatorBaseTests.cs
daniellovell: There's no test for azimuth preservation. e.g.

```cs
[Test]
public void InterceptPosition_PreservesAzimuth() {
  var targetPosition = new Vector3(70.7f, 1f, 70.7f);  // 45° azimuth
  Vector3 result = _interpolator.InterceptPosition(targetPosition);

  // Should preserve the 45° azimuth while snapping to nearest data point (100, 1)
  float expectedAzimuth = Mathf.Atan2(70.7f, 70.7f) * Mathf.Rad2Deg;
  float actualAzimuth = Mathf.Atan2(result.z, result.x) * Mathf.Rad2Deg;
  Assert.AreEqual(expectedAzimuth, actualAzimuth, 0.1f);
}
```

Assets/Scripts/Algorithms/Planning/IterativeLaunchPlanner.cs
tryuan99: you're so cracked

Assets/Scripts/Algorithms/Planning/LaunchAnglePlannerBase.cs
tryuan99: sounds good to me

Assets/Scripts/Algorithms/Planning/LaunchPlan.cs
tryuan99: Fixed in #116

daniellovell (approved):
LGTM now

PR #110, opened 2025-10-17
[Cluster] Check for terminated agents in cluster

### LINKED ISSUE #109: Carrier interceptors ditch with no target
Running the `5_swarms_100_ucav.pbtxt` configuration, I found that some carrier interceptors ditch into the sea when they have no more target.  It seems like they are heading for the origin.

Image

Image

tryuan99:
This PR resolves #109, where we noticed that carrier interceptors in the second wave or so would sometimes perform a U-tun and head back towards the origin.  The root cause is that when some threats in a cluster get destroyed by the missile interceptors of the first wave, the cluster's centroid calculation would still consider the terminated threats.  This places the centroid closer to the origin (since terminated threats are moved towards the origin), which causes the carrier interceptors of the second wave to head back.

Screenshot of `5_swarms_100_ucav.pbtxt` with no carrier interceptors turning back:
image

Example of a carrier interceptor simply ditching because its entire assigned cluster of threats has been intercepted already:
image

### DIFF
--- a/Assets/Tests/EditMode/AgglomerativeClustererTests.cs
+++ b/Assets/Tests/EditMode/AgglomerativeClustererTests.cs
@@ -8,6 +8,7 @@ using UnityEngine.TestTools;
 public class AgglomerativeClustererTests {
   public static GameObject GenerateObject(in Vector3 position) {
     GameObject obj = new GameObject();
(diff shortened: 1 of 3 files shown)


daniellovell (approved):
ship it

PR #122, opened 2025-10-26
[Interceptors/Threats] Implement all interceptor and threat types

tryuan99:
This PR fleshes out the implementations for all interceptor and threat types.  In particular, the following changes were made:
1. `CarrierBase` and `LauncherBase` abstract classes have been fleshed out.  The implementations of `CarrierInterceptor`, `MissileInterceptor`, `Vessel`, `ShoreBattery`, `FixedWingThreat`, and `RotaryWingThreat` have been added.
2. I added a `IReleaseStrategy` interface that defines when a carrier (e.g., a carrier interceptor or a launcher) releases its sub-agents.  They can be released en masse (`MassReleaseStrategyBase`) or one at a time (`SingleReleaseStrategyBase`).  Using a planner to launch sub-agents is an example of a single release strategy while releasing all interceptors in the proximity of a threat is an example of a mass release strategy.
3. `UpdateAgentConfig` is a virtual function that is called when an agent's `AgentConfig` (the dynamic configuration) changes.  This is necessary to support initializing different types of attack behaviors, evasions, controllers, and sensors.
4. The `TargetModel` should be a property of the `Agent` and not of the `Hierarchical` object, so it has been moved to the `IAgent` subclasses.  Updating the target model now happens according to the sensor's update period in the agent's `FixedUpdate` function.
5. In the simulation configuration, `submunitions_config` and `num_submunitions` have been renamed to `sub_agent_config` and `num_sub_agents`.
6. The sensor output now also returns the acceleration transformation, which is needed by the augmented proportional navigation controller.

This PR will not pass unit tests but is intended to facilitate review only.


Assets/Scripts/Interceptors/CarrierBase.cs
```diff
@@ -0,0 +1,27 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+// Base implementation of a carrier.
+//
+// A carrier carries other interceptors, such as a launcher or a carrier interceptor.
+public abstract class CarrierBase : InterceptorBase {
+  [SerializeField]
+  private int _numReleasedSubInterceptors = 0;
```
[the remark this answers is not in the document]
tryuan99: Will be used in a later PR.

Assets/Scripts/Managers/SimManager.cs
```diff
@@ -369,9 +379,42 @@ public Agent CreateDummyAgent(Vector3 position, Vector3 velocity) {
...
+    interceptor.StaticConfig = staticConfig;
+
+    // Assign a unique and simple ID.
+    int interceptorId = 0;
+    interceptorObject.name = $"{staticConfig.Name}_Interceptor_{interceptorId}";
+
+    // Add the interceptor's unit cost to the total cost.
+    _costLaunchedInterceptors += staticConfig.Cost;
+    return interceptor;
+  }
```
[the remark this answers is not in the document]
tryuan99: Will be used in a future PR!

Assets/Scripts/Agents/AgentBase.cs
```diff
@@ -71,14 +87,12 @@ public Vector3 AccelerationInput {
     set => _accelerationInput = value;
   }
 
-  private void Awake() {
-    _rigidbody = GetComponent<Rigidbody>();
+  public float ElapsedTime {
+    get { return _elapsedTime; }
+  private
+    set { _elapsedTime = value; }
```
daniellovell: nit: formatting a bit weird here with the `set` on a different line from the `private`

```csharp
public float ElapsedTime {
  get { return _elapsedTime; }
  private set { _elapsedTime = value; }
}
```

Assets/Scripts/Agents/IAgent.cs
daniellovell: Architectural concern here with the ownership of `TargetModel`

The `TargetModel` is stored in `IAgent` (via `AgentBase`) but created, destroyed, and lifecycle-managed by `HierarchicalAgent`. This is kinda split ownership?

```csharp
// HierarchicalAgent manages the lifecycle
Agent.TargetModel = SimManager.Instance.CreateDummyAgent(base.Target.Position, base.Target.Velocity);

// But AgentBase updates it
TargetModel.Position = Position + sensorOutput.Position.Cartesian;
```

unclear ownership usually leads to lifecycle bugs

What if we move TargetModel ownership entirely to HierarchicalAgent and provide a getter method in IAgent that delegates to HierarchicalAgent? Alternatively might force HierarchicalAgent to notify AgentBase when it creates/destroys target models

Assets/Scripts/Algorithms/Prediction/LinearExtrapolator.cs
```diff
@@ -8,6 +8,10 @@ public LinearExtrapolator(IHierarchical hierarchical) : base(hierarchical) {}
   // Predict the future state of the hierarchical object by linearly extrapolating its current
   // position and velocity.
   public override PredictorState Predict(float time) {
+    if (Hierarchical == null) {
+      return new PredictorState();
+    }
+
```
daniellovell: Returning a default/empty `PredictorState` seems misleading since callers cannot distinguish between "prediction failed" and "predicted position is (0,0,0)"

Recommend throwing an exception or returning null and updating the callers to handle null. At the least log a warning/error.

Assets/Scripts/Agents/IAgent.cs
```diff
@@ -13,18 +13,23 @@ public interface IAgent {
   IMovement Movement { get; set; }
   IController Controller { get; set; }
   ISensor Sensor { get; set; }
+  IAgent TargetModel { get; set; }
```
daniellovell: Documentation needed
```suggestion
  // Target model maintained by the agent's sensor. Represents the agent's best estimate
  // of the target's state based on sensor measurements and prediction.
  // Managed by HierarchicalAgent when targets are assigned/unassigned.
  IAgent TargetModel { get; set; }
```

daniellovell (approved):
As far as I can tell, only changes since last time reviewed are addressing comments (these LGTM) + merging. 

Send it

PR #129, opened 2025-10-26
[Release] Adjust minimum distance in proximity release

tryuan99:
Previously, carrier interceptors would release their submunitions if the target was predicted to be within 500 meters of the carrier.  However, I found an error in the submunitions release that caused the submunitions to have unexpectedly good intercept performance.  Their velocity was set twice, once by `SimManager`'s `CreateInterceptor` function and another time erroneously by directly setting the missile interceptor's velocity:
https://github.com/PisterLab/micromissiles-unity/blob/a6f7de422991e6abf739f6238a55ff8a34bf69fc/Assets/Scripts/Interceptors/CarrierInterceptor.cs#L62
This caused the missile interceptors to accelerate to up to 1.5 km/s, giving them better agility than expected.

After fixing this bug, the missile interceptors now only accelerate up to around 1.2 km/s, so they need more maneuvering distance to be able to intercept the threats.  This PR thus sets the minimum release distance to 1 km instead of 500 m.

Example of releasing at a minimum distance of 500 meters:
image

To allow interceptors to also be able to intercept supersonic threats, I increased the maximum release distance to 2500 meters.


daniellovell (approved):
LGTM

You said
> To allow interceptors to also be able to intercept supersonic threats, I increased the maximum release distance to 2500 meters.

Do we have a threat to validate this or is it more of an intuition thing?

PR #132, opened 2025-10-27
[Simulation] Destroy terminated game objects

tryuan99:
This PR changes the simulation manager to directly destroy game objects when the corresponding agents have been terminated instead of deactivating them.  One required change for this is that hierarchical objects only deal with active sub-hierarchical objects to prevent throwing `MissingReferenceException`s.

This change was tested in the Unity IDE by restarting the simulation multiple times to ensure that the simulation manager can deal with agents that have already been destroyed.  Furthermore, when a carrier interceptor assigns missile interceptors to the cluster's sub-hierarchical objects, only the active ones are considered.


daniellovell (approved):
Looks like review was fully addressed awhile ago, sorry

PR #133, opened 2025-10-27
[UI] Draw gizmos for agents

tryuan99:
For debugging, draw gizmos showing the acceleration input for all agents and draw the three axes of rotation for all interceptors.  To reduce clutter, the axes of rotation are not drawn for threats.

Examples:
Acceleration vectors on threats:
image

Acceleration vectors and lines to threats on interceptors:
image

Axes of rotation within an interceptor:
image


daniellovell (approved):
Works on my machine and looks sick 🤙 

Image

PR #134, opened 2025-10-27
[UI] Fix summary text label sizes

tryuan99:
As title.

Before:
image

After:
image


daniellovell (approved):
Was not able to reproduce the CI build failure on Windows. Builds fine.

Text fix works machine but I'm noting two small bugs here, I'll make issues for them. Or actually, @coderabbitai can you make two issues for these (adopt our Issue style in title and description) - Edit: Lol he ignored me.

Image

1. The number of Threats Remaining can somehow show as -1 in the UI
2. The buttons in the center of the UI used to be centered I believe in one single column, now the Clear Markers button is offset from the other three.

PR #139, opened 2025-11-06
[Clustering] Implement recursive clustering within hierarchy

tryuan99:
This PR implements recursive clustering within the hierarchical objects.  Recursive clustering starts at the IADS, which performs a _k_-means clustering algorithm to assign a threat swarm to each launcher.  Afterwards, each hierarchical object will cluster the incoming collection of threats and create sub-hierarchical objects correspondingly.

This PR also fixes the mass release strategy to check against each sub-threat in the target hierarchical to see whether all sub-interceptors should be released.  This is necessary, so that a carrier interceptor does not release its sub-interceptors at the cluster centroid, which may be behind some threats.

The agent hierarchy looks something as follows:
image


Assets/Scripts/IADS/IADS.cs
```diff
@@ -7,13 +7,12 @@
 public class IADS : MonoBehaviour {
   // Clustering parameters.
   private const float _clusterThreatsPeriod = 5f;
-  private const int _clusterMaxSize = 7;
-  private const float _clusterMaxRadius = 500f;
+  private const float _coverageFactor = 1f;
```
daniellovell: Needs documentation

```suggestion
  // Factor controlling cluster-to-launcher ratio.
  // 1.0 = one cluster per launcher (default)
  // 0.5 = two clusters per launcher (redundant coverage)
  // 2.0 = one cluster for every two launchers (sparse coverage)
  private const float _coverageFactor = 1f;
```

Assets/Scripts/Release/MassReleaseStrategyBase.cs
```diff
@@ -16,62 +16,67 @@ public MassReleaseStrategyBase(IAgent agent, IAssignment assignment) : base(agen
...
+        Velocity = velocityCoordinates,
+      };
+      IAgent subInterceptor =
+          SimManager.Instance.CreateInterceptor(subAgentConfig.AgentConfig, initialState);
+      if (subInterceptor != null && subInterceptor is IInterceptor subInterceptorInterceptor) {
+        if (subInterceptor.Movement is MissileMovement movement) {
+          movement.FlightPhase = Simulation.FlightPhase.Boost;
         }
+        releasedAgents.Add(subInterceptor);
       }
```
daniellovell: Why does `subInterceptorInterceptor` declared and type checked here? Was it meant to be used? Right now its declared but ignored. Confusing, maybe some incomplete refactoring

Assets/Scripts/Hierarchical/HierarchicalBase.cs
```diff
@@ -67,7 +64,60 @@ public void RemovePursuer(IHierarchical pursuer) {
...
+      clusterer = new AgglomerativeClusterer(maxClusterSize, _clusterMaxRadius);
+    }
+    List<Cluster> clusters = clusterer.Cluster(Target.SubHierarchicals);
+
+    // Generate sub-hierarchical objects to manage the target clusters.
+    foreach (var cluster in clusters) {
+      var subHierarchical = new HierarchicalBase { Target = cluster };
+      AddSubHierarchical(subHierarchical);
+      subHierarchical.RecursiveCluster(maxClusterSize);
+    }
```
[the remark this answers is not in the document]
tryuan99: No, it's not possible to iterate over a collection that is changing in this function.  It's recursive down the tree, so:
1) if the node is a leaf, create sub-clusters and call `RecursiveCluster` on each newly created child.
2) if the node is not a leaf, call `RecursiveCluster` on each child and return directly (so the current node's `SubHierarchicals` list is unchanged).

Assets/Scripts/IADS/IADS.cs
tryuan99: IADS is dead code in this PR.

Assets/Scripts/Release/MassReleaseStrategyBase.cs
tryuan99: Good catch.

daniellovell (approved):
changes LGTM

PR #142, opened 2025-11-06
[Gemini] Disable Gemini pull request summary

tryuan99:
Disable Gemini from posting a pull request summary whenever a pull request is opened.  The summary should be in the description box of the PR and should be ideally human-written for conciseness and helpfulness.

https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github#config.yaml-example for how to configure Gemini for Github.

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,0 +1,11 @@
+have_fun: false
+code_review:
+  disable: false
+  comment_severity_threshold: MEDIUM
+  max_review_comments: -1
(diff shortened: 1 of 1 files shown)


.gemini/config.yaml
```diff
@@ -0,0 +1,11 @@
+have_fun: false
```
daniellovell: 🥲 fun is cancelled

daniellovell (approved):

PR #143, opened 2025-11-06
[Evasion] Fix threat evasion of interceptors

tryuan99:
Due to the agent hierarchy, when an interceptor is assigned a target, it is usually assigned a cluster of targets (of which there may only be one in the case of a missile interceptor).  To set the pursuers correctly, we thus traverse the target hierarchy and add the interceptor as a pursuer to all targets.


daniellovell (approved):
LGTM

PR #144, opened 2025-11-06
[Reassignment] Use capacity remaining to reassign interceptors

tryuan99:
As title.  The capacity remaining may be less than the capacity for carriers if they have already released some sub-interceptors.


daniellovell (approved):
LGTM

PR #147, opened 2025-11-06
[Escape] Add escape detectors

tryuan99:
This PR implements the escape detector interface and its subclass implementations.  The escape detectors determine heuristically whether a target is escaping its pursuer.  This PR implements three different escape detectors:
1. The geometric escape detector checks whether the agent is moving head-on towards the target.
2. The time-based escape detector checks whether the target will hit before the agent hits the target.
3. The speed-based escape detector checks whether the agent has a greater speed than the target when the agent reaches the target.

Unit tests are provided for all three escape detectors.


Assets/Scripts/Escape/TimeEscapeDetector.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    // Check if the target has a target of its own.
+    if (target.Target != null && !target.Target.IsTerminated) {
+      Vector3 targetRelativePositionToTarget = target.Target.Position - target.Position;
+      // Check whether the agent is moving head-on towards the target.
+      if (Vector3.Dot(-agentToTargetTransformation.Position.Cartesian,
+                      targetRelativePositionToTarget) > 0) {
+        // Check the time-to-hit for the agent and the target.
+        float targetTimeToHit = targetRelativePositionToTarget.magnitude / target.Speed;
+        float agentTimeToHit = agentToTargetTransformation.Position.Range / Agent.Speed;
```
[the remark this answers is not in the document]
daniellovell: Suggest:
```csharp
if (target.Speed < 1e-6f) {
  return false; // Or appropriate fallback logic
}
float targetTimeToHit = targetRelativePositionToTarget.magnitude / target.Speed;
```

Assets/Scripts/Escape/EscapeDetectorBase.cs
```diff
@@ -0,0 +1,14 @@
+using UnityEngine;
+
+// Base implementation for an escape detector.
+public abstract class EscapeDetectorBase : IEscapeDetector {
+  // Agent that is pursuing the target.
+  public IAgent Agent { get; init; }
+
+  public EscapeDetectorBase(IAgent agent) {
+    Agent = agent;
+  }
```
daniellovell: All three escape detectors check for null target but dont check if `Agent` is null or the `Agent.StaticConfig` is null.
Low hanging fruit null check might be here with
```csharp
public EscapeDetectorBase(IAgent agent) {
  if (agent == null) {
    throw new ArgumentNullException(nameof(agent));
  }
  Agent = agent;
}
```

Assets/Scripts/Escape/GeometricEscapeDetector.cs
```diff
@@ -0,0 +1,35 @@
...
+      Vector3 targetRelativePositionToTarget = target.Target.Position - target.Position;
+      // The target is escaping if the distance from the agent to the target is greater than the
+      // distance from the target to its target or if the agent is geometrically behind the target
+      // relative the target's own target.
+      return agentToTargetTransformation.Position.Range >
+                 targetRelativePositionToTarget.magnitude * _rangeBufferFactor ||
+             Vector3.Dot(-agentToTargetTransformation.Position.Cartesian,
+                         targetRelativePositionToTarget) < 0;
+    }
+
```
daniellovell: There's kind of an inconsistency in behavior here where both detectors fall back to checking `agentToTargetTransformation.Velocity.Range < 0` when the target has no target, but they check this condition differently.

The geometric escape detector can declare escape even with positive closing velocity if the agent is geometrically behind, but the time escape detector doesn't have this opportunity/behavior? Maybe this is intentional? Could use explanation in comments maybe

Assets/Scripts/Escape/EscapeDetectorBase.cs
tryuan99: I'd even say that a precondition is that `agent` is non-null.  Otherwise, throwing the exception here is equivalent to a NRE being thrown in the code.  All the hierarchical algorithms actually assume a non-null Agent - if you have any ideas on how to take care of all of them checking for non-null, that might be a cleaner idea (than a null check everywhere).

Assets/Scripts/Escape/GeometricEscapeDetector.cs
tryuan99: Yes, this is intentional.  The time escape detector only cares about the time (at least this was the thought in my mind).  I don't think any of these escape detectors are necessarily optimal, just ideas that I had.

Assets/Tests/EditMode/GeometricEscapeDetectorTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+  [Test]
+  public void IsEscaping_AgentOrthogonalToTarget_ReturnsFalse() {
+    _agent.Position = new Vector3(0, 1, 0);
+    var target =
+        new FixedHierarchical(position: new Vector3(0, 0, 10), velocity: new Vector3(0, 0, -1)) {
+          Target = new FixedHierarchical(position: new Vector3(1, 0, 0))
+        };
+    Assert.IsFalse(_escapeDetector.IsEscaping(target));
+  }
```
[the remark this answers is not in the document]
tryuan99: Fixed, thanks!

daniellovell (approved):

PR #148, opened 2025-11-06
[Escape] Check for escaping targets

tryuan99:
This PR uses the previously implemented escape detectors to implement escape detection on the interceptors.  When an interceptor detects that its target is escaping, the interceptor reports the escaping target up the hierarchy (target reassignment) and, if no other targets remain for the interceptor, the interceptor requests a new target from the parent interceptor.

Carrier interceptors use a geometric escape detector since they must release prior to reaching their target cluster centroid, and missile interceptors use a time-based escape detector.

Another change in this PR is that interceptors that miss their targets are not destroyed, so that they can circle back to try to intercept their targets again.


Assets/Scripts/Interceptors/InterceptorBase.cs
```diff
@@ -69,19 +71,35 @@ protected override void Start() {
...
+    // Check whether any targets are escaping from the interceptor.
+    if (EscapeDetector != null && HierarchicalAgent.Target != null &&
+        !HierarchicalAgent.Target.IsTerminated) {
+      List<IHierarchical> targetHierarchicals =
+          HierarchicalAgent.Target.LeafHierarchicals(activeOnly: true, withTargetOnly: false);
+      List<IHierarchical> escapingTargets =
+          targetHierarchicals.Where(EscapeDetector.IsEscaping).ToList();
+      foreach (var target in escapingTargets) {
+        OnReassignTarget?.Invoke(target);
+      }
```
daniellovell: I cant find where the current interceptor is removed from the target's pursuer list. Shouldn't `target.RemovePursuer(HierarchicalAgent)` be added?

Same in `RegisterMiss`?:
```csharp
foreach (var targetHierarchical in targetHierarchicals) {
  targetHierarchical.RemovePursuer(HierarchicalAgent);
  OnReassignTarget?.Invoke(targetHierarchical);
}
```

tryuan99: It's called when the hierarchical's `Target` changes:
```cs
      if (base.Target != null) {
        base.Target.RemovePursuer(this);
        ClearSubHierarchicals();
        SimManager.Instance.DestroyDummyAgent(Agent.TargetModel);
        Agent.TargetModel = null;
      }
```

Assets/Scripts/Interceptors/InterceptorBase.cs
```diff
@@ -69,19 +71,35 @@ protected override void Start() {
...
+      if (escapingTargets.Count == targetHierarchicals.Count) {
+        OnAssignSubInterceptor?.Invoke(this);
+      }
```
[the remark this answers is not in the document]
tryuan99: This will be done when the new target is assigned.  Doing it earlier is problematic if there are no other targets to assign to.

daniellovell (approved):
Should have approved this with comments originally, sorry.

PR #150, opened 2025-11-07
[Reassignment] Fix relaunching multiple carrier interceptors

tryuan99:
There was a bug in the previous implementation, where since each target only tracked the missile interceptors as pursuers (and no carriers), the launcher might fire multiple carriers at an escaping target.  The constant relaunching would only cease once a carrier deployed its missile interceptors as the target would then have non-escaping pursuers.

This PR fixes that bug by having the targets track the carriers as pursuers as well; only launchers are considered non-pursuable.  A launcher will only fire at an escaping target if the target is escaping all of its pursuers or the non-escaping pursuers have zero capacity remaining (as is the case with a carrier that has dispensed all of its submunitions).


daniellovell (approved):
LGTM

PR #152, opened 2025-11-07
[Unity] Fix serializable fields and references

tryuan99:
This PR fixes setting `[SerializeField]` and `[SerializeReference]` within the agent hierarchy and hierarchical objects to ensure that the Unity Inspector can actually display the serialized fields.

This is the current inspector view of a carrier interceptor:
image

This PR also enlarges the launchers to be around as large as the carrier interceptors for better visibility.
image


daniellovell (approved):
LGTM

PR #157, opened 2025-11-16
[Visualization] Fix visualizing simulation monitor logs

tryuan99:
This PR fixes visualizing the simulation monitor logs.  In particular, I revamped the `visualize_log.py` script with the Abseil library and proper docstrings for all functions.  I also fixed the session directory naming in `SimMonitor.cs`, which was using backslashes in the log file path instead of using `Path.Combine`.  Finally, I properly destroy the logging in `SimMonitor.cs` for both telemetry logging and event logging.

Example output of the log visualization script:
image


.gitignore
```diff
@@ -73,16 +73,16 @@ crashlytics-build.properties
 .vscode/
 .vsconfig
```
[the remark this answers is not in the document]
tryuan99: This is not in the repo path though.

Tools/visualize_log.py
```diff
@@ -1,221 +1,211 @@
...
+    ax.set_ylabel("$z$ [m]")
+    ax.set_zlabel("$y$ [m]")
 
-def print_summary(telemetry_file_path, event_file_path):
-    # Read the telemetry CSV file.
-    df = pd.read_csv(telemetry_file_path)
+    ax.set_aspect("equal")
+    ax.view_init(elev=20, azim=-45)
+    fig.tight_layout()
+    plt.show()
```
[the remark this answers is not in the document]
tryuan99: No.

Tools/visualize_log.py
```diff
@@ -1,221 +1,211 @@
...
-        )
-    elif platform.system() == "Darwin":
+            r"%USERPROFILE%\AppData\LocalLow\BAMLAB\micromissiles\Logs")
+    elif system == "Darwin":
         return os.path.expanduser(
-            "~/Library/Application Support/BAMLAB/micromissiles/Telemetry/Logs"
-        )
+            "~/Library/Application Support/BAMLAB/micromissiles/Logs")
+    elif system == "Linux":
+        return os.path.expanduser("~/.config/unity3d/Logs")
```
daniellovell: I feel like this is wrong? Shouldn't it be

```suggestion
        return os.path.expanduser("~/.config/unity3d/BAMLAB/micromissiles/Logs")
```

Hunch tho

Tools/visualize_log.py
```diff
@@ -1,221 +1,211 @@
...
+    }
 
-    # Group data by AgentID.
-    agent_types = set()
-    for agent_id, agent_data in df.groupby("AgentID"):
+    # Plot the agent trajectories.
+    for _, agent_data in telemetry_df.groupby("AgentID"):
         agent_type = agent_data["AgentType"].iloc[0]
         color = colors.get(agent_type, "black")
-        downsampled = agent_data.iloc[::10]
```
daniellovell: Used to be downsampled cause there would be way too much data and the 3d viz would be laggy. Double checking that its intentional to remove data downsample.

Tools/visualize_log.py
```diff
@@ -1,221 +1,211 @@
...
-    # Optimize legend to prevent overcrowding.
-    handles, labels = ax.get_legend_handles_labels()
-    # Remove duplicate labels.
-    unique = dict(zip(labels, handles))
-    ax.legend(unique.values(),
-              unique.keys(),
-              loc="upper left",
-              bbox_to_anchor=(1, 1),
-              fontsize=10)
-
```
daniellovell: No more legend :(

Tools/visualize_log.py
```diff
@@ -1,221 +1,211 @@
...
+    # Calculate the time duration of the events.
+    start_time = event_df["Time"].min()
+    end_time = event_df["Time"].max()
+    duration = end_time - start_time
+    logging.info("Total duration of events: %.2f seconds (from %.2f to %.2f).",
+                 duration, start_time, end_time)
+
+    # Determine the times of the hits and misses.
+    hits = event_df[event_df["Event"] == "INTERCEPTOR_HIT"]
+    misses = event_df[event_df["Event"] == "INTERCEPTOR_MISS"]
```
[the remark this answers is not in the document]
tryuan99: I don't want to support the old log, I don't think there's anything we have from there anyway worth keeping (no hierarchical stuff either)

Tools/visualize_log.py
tryuan99: Fixed in #160

Tools/visualize_log.py
tryuan99: Yes, I was able to successfully plot with 2_ddg_5_swarms_100_ucav.pbtxt without any issues.  Also the simulator only logs telemetry every 0.1 seconds.

Tools/visualize_log.py
tryuan99: Fixed.  Looks like this now:

Image

daniellovell (approved):

PR #158, opened 2025-11-16
[Tools] Build standalone player

tryuan99:
This PR adds a script to build a standalone Unity player and is adapted from #67.  I refactored all Unity-related functionalities, including finding the Unity installations in the system, into a separate file `unity.py`.  This PR was tested by running the generated standalone player and verifying a few of the simulation configurations.


Tools/unity.py
daniellovell: nit: unity_utils.py or unity_utilities.py would convey the purpose of this file at a glance

Tools/build_standalone.py
```diff
@@ -0,0 +1,102 @@
...
+# Default name of the standalone build.
+DEFAULT_BUILD_NAME = "micromissiles"
+
+
+@dataclass(frozen=True)
+class PlatformTarget:
+    """Platform target."""
+    unity_name: str
+    build_flag: str
+    extension: str
```
daniellovell: add detailed docstring
```suggestion
@dataclass(frozen=True)
class PlatformTarget:
    """Platform-specific build target configuration.

    Attributes:
        unity_name: Unity's internal platform name (e.g., "StandaloneWindows64").
        build_flag: Command-line flag for building this platform.
        extension: File extension for the built executable (empty string for Linux).
    """
```

daniellovell (approved):

daniellovell (approved):

tigran-sargsyan-w (commented):
```markdown
Nice PR, I really like the direction here.

`build_standalone.py` is easy to follow and the timestamped build folders under `Build/` are super handy for keeping runs separate. Extracting the Unity discovery logic into `unity_utils.py` also makes the tooling a lot easier to reuse and test in isolation.

A couple of tiny thoughts from a drive-by review:
- `PlatformTarget` is already quite clear, but a slightly more descriptive docstring (what `unity_name` / `build_flag` / `extension` mean) could help future contributors who don’t know the Unity CLI by heart.
- In `find_unity_path`, when multiple installs are found you log a warning and pick the first one. It might be useful to either log the full list or allow selecting a version via a flag/env var in the future.

Overall this looks solid and pragmatic, and the defaults (env var + auto-discovery) feel very reasonable.
```

PR #160, opened 2025-11-17
[Tools] Refactor common simulation-related constants and utilities

tryuan99:
This PR refactors common simulation-related constants  and utilities into separate files for better reusability.  In particular, all simulation-related constants, including the agent types, the event types, and the log file prefixes, are now defined in `constants.py` while utility functions, including finding the latest telemetry file and event log, are now defined in `utils.py`.


daniellovell (approved):
LGTM 👍

PR #167, opened 2025-11-18
[Agent] Add target model API in Agent interface

tryuan99:
As title.  Addresses https://github.com/PisterLab/micromissiles-unity/pull/122#discussion_r2536697626 and https://github.com/PisterLab/micromissiles-unity/pull/122#discussion_r2536713948.


Assets/Scripts/Agents/IAgent.cs
```diff
@@ -36,6 +45,8 @@ public interface IAgent {
   float MaxForwardAcceleration();
   float MaxNormalAcceleration();
 
+  void CreateTargetModel(IHierarchical target);
+  void DestroyTargetModel();
```
daniellovell: Need documentation for the new methods. Unclear when/how to call them, who should call them, or relationship to TargetModel property.


```suggestion
  // Create a dummy agent to represent the target for sensor tracking and   control.
  // Should be called when a target is assigned to this agent.
  // The created dummy agent will be stored in TargetModel.
  void CreateTargetModel(IHierarchical target);
  
  // Destroy the target model dummy agent and clear the TargetModel   reference.
  // Should be called when the target is unassigned or no longer valid.
  void DestroyTargetModel();
```

daniellovell (approved):

PR #169, opened 2025-11-19
[Metrics] Consider only missile interceptors in metrics

tryuan99:
This PR changes the default metrics to only consider missile interceptors.  Currently, the metrics are confounded by launchers and carriers being created and missing (when the carrier interceptors hit the ground).  Thanks to the agent type in the event logs, we can easily filter for the `MissileInterceptor` agent type.


daniellovell (approved):
LGTM

PR #171, opened 2025-11-19
[Logging] Remove logging from CP-SAT solver

tryuan99:
As title.  Logging the progress of the CP-SAT solver blows up the Editor log to 10s of GBs for <50 runs.


daniellovell (approved):
Thank you for sparing my file system

PR #172, opened 2025-11-20
[Tools] Add script to animate telemetry log

tryuan99:
This PR adds a script to generate an HTML file containing an interactive 3D animation of the telemetry file.  The output looks as follows:
image

There could be many more improvements to enhance the appearance of the telemetry animation, but I am not a frontend engineer.


daniellovell (approved):
This is sick! What do you think about `replay_log` instead of `animate_log`?

PR #177, opened 2025-12-07
[UI] Add camera to follow and rotate around interceptors or threats

tryuan99:
This PR changes the camera controller to allow (1) following and snapping to all interceptors or all threats in addition to all agents and 2) orbiting the camera around a desired position.  I decreased the rotation camera speed significantly to avoid vertigo, but it could be further tuned (and should be reviewed please).

anim


daniellovell (approved):
So cracked

PR #179, opened 2025-12-08
[Configs] Add configuration for 5 swarms and 1000 UCAVs

tryuan99:
This PR adds a simulation configuration for 5 swarms with 1000 UCAVs total and adds run configurations for the `5_swarms_500_ucav.pbtxt` and `5_swarms_1000_ucav.pbtxt` configs.


daniellovell (approved):
LGTM. Reminds me will need to debug the issue with batch runs on Windows which I saw when running these configs, but that's not in the scope of this PR

PR #180, opened 2025-12-09
[Agent] Cache the agent transform

tryuan99:
This PR attempts to reduce the number of calls to `transform`, including getting the `position`, `up`, `forward`, and `right`.  This PR caches all of these values within `AgentBase` and adds an `EarlyFixedUpdateManager` to refresh all cache values prior to all agents' `FixedUpdate` calls.  To ensure that `EarlyFixedUpdateManager` is executed before the agents, its execution order has been set to `-50`.

Profiler before change:
image

Profiler after change:
image


Assets/Scripts/Agents/AgentBase.cs
```diff
@@ -155,7 +174,13 @@ public void Terminate() {
...
   // Awake is called before Start and right after a prefab is instantiated.
   protected virtual void Awake() {
+    Transform = transform;
     _rigidbody = GetComponent<Rigidbody>();
+
+    UpdateTransformData();
+    if (EarlyFixedUpdateManager.Instance != null) {
+      EarlyFixedUpdateManager.Instance.OnEarlyFixedUpdate += UpdateTransformData;
+    }
   }
```
daniellovell: One minor concern here - if `AgentBase.Awake` runs before `EarlyFixedUpdateManager.Awake`, the subscription should fail (and may do so silently). Unity doesn't guarantee `Awake()` ordering across different objects (only script execution order affects Update/FixedUpdate). May cayse non-deterministic behavior depending on scene initialization order?

Can come back to this in the future or just make a mental note in case problems are seen in the future, we may know how to fix quickly

daniellovell (approved):

PR #181, opened 2025-12-09
[Agent] Clean up finding relative transformations

tryuan99:
From the profiler (shown below), calculating the relative transformation between agents uses the most resources.  This PR changes `GetRelativeTransformation` to use quaternions and inverse rotations instead of `Vector3.Dot` and `Vector3.SignedAngle`.

Profiler before:
image

Profiler after:
image


Assets/Tests/EditMode/PnControllerTests.cs
```diff
@@ -97,11 +97,11 @@ public void Plan_TargetOverhead_MovingInParallel() {
...
-    // = 1.
+    // Horizontal acceleration = gain * closing velocity * azimuth line-of-sight rate = 1 * 1 * 0.2
+    // = 0.2.
     // Vertical acceleration is clamped, so gain * closing velocity * elevation line-of-sight rate =
-    // 1 * 1 * 0.2 = 0.2. The acceleration is multiplied by a turn factor of 100 to force a quicker
-    // turn.
-    Assert.AreEqual(new Vector3(1, 0.2f, 0) * 100, _controller.Plan());
+    // 1 * 1 * -1 = -1.
+    // The acceleration is multiplied by a turn factor of 100 to force a quicker turn.
+    Assert.AreEqual(new Vector3(0.2f, -1, 0) * 100, _controller.Plan());
```
tryuan99: This unit test changed if the target is above me and moving to the right, I think it should count more as an elevation rate (negative) than an azimuth rate because the direction in which the azimuth is changing is ambiguous.

Assets/Tests/EditMode/PnControllerTests.cs
```diff
@@ -97,11 +97,11 @@ public void Plan_TargetOverhead_MovingInParallel() {
...
   }
```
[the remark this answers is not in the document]
tryuan99: no

Assets/Scripts/Agents/AgentBase.cs
```diff
@@ -256,109 +260,60 @@ private void AlignWithVelocity() {
...
+      elevationRate =
+          (relativeLocalVelocity.y * horizontal -
+           y * (x * relativeLocalVelocity.x + z * relativeLocalVelocity.z) / horizontal) /
+          rangeSqr;
+    } else {
+      // The other agent is exactly above or below.
+      azimuthRate = 0f;
+      float horizontalSpeed = Mathf.Sqrt(relativeLocalVelocity.x * relativeLocalVelocity.x +
+                                         relativeLocalVelocity.z * relativeLocalVelocity.z);
+      elevationRate = -horizontalSpeed / (Mathf.Abs(y) > _epsilon ? y : Mathf.Sign(y) * _epsilon);
```
daniellovell: Potential divide-by-zero here since Mathf.Sign(0) returns 0 so if y is 0 (exactly at same alt), this becomes div-by-zero.

daniellovell (approved):
ship it 

If we care in future maybe useful to add unit tests for singularity cases like:
 - Target overhead moving horizontally
 - Target underfoot moving horizontally
 - Target overhead moving down and to the side
 - Target at exact same altitude (y = 0)

PR #182, opened 2025-12-16
[Configs] Exclude 5_swarms_1000_ucav configuration from CI

tryuan99:
As title.  The CI runner gets stuck trying to cluster all 1000 UCAVs even though it runs within a few seconds on my Mac.

Example: https://github.com/PisterLab/micromissiles-unity/actions/runs/20253233631/job/58150025987
image


daniellovell (commented):
Maybe instead of removing it, we should support some way to omit specific Sim Configs from the CI?
I can take a stab at this real quick

tryuan99:
I added a (temporary) solution to exclude `5_swarms_1000_ucav.pbtxt` from `ConfigsTests`, but it excludes that configuration file everywhere instead of only on CI.

daniellovell (approved):

PR #184, opened 2025-12-16
[Agents] Fix collisions with terminated agents

### LINKED ISSUE #135: [UI] Threats Remaining can display negative value (-1)
The Threats Remaining counter in the UI can somehow display a value of -1, which is incorrect since the number of remaining threats should never be negative.

This issue was discovered during review of PR #134.

Reported by: @daniellovell
Reference: https://github.com/PisterLab/micromissiles-unity/pull/134#discussion_r1901058008

tryuan99:
This PR resolves #135.  The root cause is that when multiple interceptors hit the same threat, one of them hits the threat and destroys it, but the others also trigger collisions due to their proximity to the (destroyed) threat.  This PR adds a check for whether the intercepted threat is destroyed before tallying it as an additional hit.

image


daniellovell (approved):
Nice, lgtm

PR #187, opened 2025-12-17
[Reassignment] Find subhierarchicals with at least one active target

### LINKED ISSUE #185: [Reassignment] Reassignment is not reassigning interceptors to next swarm
Image

This can be reproduced by setting the kill probability to 1.

tryuan99:
This PR resolves #185.  The root cause is that when reassigning interceptors, we search for all subhierarchicals that have no more active targets than the interceptor capacity, which includes subhierarchicals with zero active targets (i.e., the targets have already been terminated).  This causes the interceptors to then head back to the origin to intercept the terminated threats.  This PR fixes this issue by ensuring that only sub-hierarchicals with at least one active target are considered.

Example of `5_swarms_100_ucav.pbtxt`, where missile interceptors are continuously reassigned as they fly through all five swarms and only head back once all threats are behind them:
image


daniellovell (approved):
This makes sense to me, looks good 👍

PR #188, opened 2025-12-17
Restore 5_swarms_1000_ucav configs & add CI skip list

daniellovell:
Restore the removed `5_swarms_1000_ucav` simulation + batch run configs, but keep CI stable by letting the PlayMode config-loading test skip specific heavy configs when `CI=true/1`.

- Restores `Assets/StreamingAssets/Configs/Simulations/5_swarms_1000_ucav.pbtxt` and `Assets/StreamingAssets/Configs/Runs/batch_5_swarms_1000_ucav.pbtxt`
- Adds `Assets/StreamingAssets/Configs/Simulations/ci_excluded_simulation_configs.txt` (one filename per line)
- Updates `Assets/Tests/PlayMode/ConfigsTests.cs` to consult the exclusion list only in CI

### DIFF
--- a/Assets/Tests/PlayMode/ConfigsTests.cs
+++ b/Assets/Tests/PlayMode/ConfigsTests.cs
@@ -1,5 +1,5 @@
-using System;
 using NUnit.Framework;
+using System;
 using System;
 using System.IO;
 using System.Collections;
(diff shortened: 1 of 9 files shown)


Assets/Tests/PlayMode/ConfigsTests.cs
```diff
@@ -1,21 +1,43 @@
+using System;
```
tryuan99: nit: alphabetic include order

tryuan99 (approved):
Thanks!

tryuan99:
Looks like the latest change works:

image

image

PR #200, opened 2026-01-30
[Reassignment] Interceptor reassignment optimizes for intercept speed

### LINKED ISSUE #197: [Reassignment] Interceptor reassignment tends to only target one threat
When interceptor reassignment occurs for a group of interceptors, they all tend to pursue the same threat, which might not be optimal.  Debug how the reassignment strategy is done and it might perhaps make sense to distribute the reassigned interceptors.

tryuan99:
This PR resolves #197.  The problem with the previous interceptor reassignment occurs when a missile interceptor with capacity 1 is reassigned to a new threat that is selected from a leaf hierarchical's target cluster.  Previously, this selection was based on the distance from the threat's position to the cluster's centroid (so all interceptors were assigned to the same threat).  This PR changes this selection process to optimize for the intercept speed.

To clean up the code, calculating the fractional speed was refactored into its standalone static class `FractionalSpeed`.  Furthermore, I modified the `AssignNewTarget` method to `FindNewTarget` to return a candidate threat.  Each interceptor will then decide whether to accept the new threat or not in the `EvaluateReassignedTarget` method.

In this screenshot, you can see that most interceptors are reassigned to the next swarm of threats instead of forced to turn around to deal with the single missed threat from the first swarm.  The reassigned targets are also spread out depending on the interceptor locations.

image

cc: @Joseph0120

### DIFF
--- a/Assets/Scripts/Agents/AgentBase.cs
+++ b/Assets/Scripts/Agents/AgentBase.cs
@@ -115,7 +115,9 @@ public class AgentBase : MonoBehaviour, IAgent {
 
   public float MaxNormalAcceleration() {
     float maxReferenceNormalAcceleration =
-        (StaticConfig.AccelerationConfig?.MaxReferenceNormalAcceleration ?? 0) * Constants.kGravity;
+        (StaticConfig.AccelerationConfig?.MaxReferenceNormalAcceleration ??
+         float.PositiveInfinity) *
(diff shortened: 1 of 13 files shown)


daniellovell (approved):
Resulting improvement to reassignment looks sick, and the logic of optimizing on intercept speed makes great sense to me 👍

PR #201, opened 2026-01-30
[Configs] Add asset configuration

### LINKED ISSUE #195: [Agent] Define an asset class
The asset class represents the target of the threats.  If there are multiple assets, threats should prioritize the closest asset.

tryuan99:
This PR adds an additional field to the simulation configuration for asset configurations.  Since I have to regenerate all plugins (as the proto files changed), I figured I might as well bump all package versions.


daniellovell (approved):
LGTM when tested (although I can't see the new assets without #202 ofc)

Curious if you see the 2_shore_batteries... config wrapping poorly in the listbox on Mac?
Image

PR #202, opened 2026-01-30
[IADS] Visualize and assign threats to target assets

### LINKED ISSUE #195: [Agent] Define an asset class

tryuan99:
This PR visualizes the assets and changes the IADS to assign the threats to target the closest asset.  Assets are drawn with an orange color in the 3D view:

image


daniellovell (commented):
I think we were discussing this a bit on Friday together - assets are also launchers right? e.g. a DDG is both an asset to be protected and a launcher? As well as a CVN is an asset and a launcher? IMO the taxonomy should be such that all launchers are assets, but not all assets are launchers. 

A very common mission for a DDG would be to escort a civilian/commercial vessel. In that case there would be two assets: the DDG and the commercial vessel, but one launcher.

Thoughts on this?

Also maybe a higher-level consideration out of the scope of this PR. You mentioned that the IADS is doing assignment for threats. That seems counterintuitive to me - the IADS (as an integrated air _defense_ system) should be responsible solely for coordinating BLUFOR's defense strategy. It feels like a separate entity from the IADS should be responsible for coordinating the threat's assignment and attack strategy. That way there is clearly a "BLUFOR coordinator" and "OPFOR coordinator"

daniellovell (approved):
Sweet! LGTM 👍

PR #205, opened 2026-01-30
[Logging] Check for null agents using MonoBehaviour

tryuan99:
Resolves #203.  During the simulation destruction, agents might already be destroyed, but the `SimMonitor` still wants to log one last telemetry data point for them.  `IAgent` is an interface independent of `MonoBehaviour`, so the null check does not return correctly for fake-null objects.  Instead, the agent has to be cast to `AgentBase`, which will then invoke the fake-null-compatible comparison with `null`.


daniellovell (approved):
Nice lgtm 👍

PR #213, opened 2026-04-01
[Communication] PR2: Adds CommsManager and CommsNodes

Joseph0120:
Files Modified:

IadsCommsAgent.cs -> IADS Proxy for supporting mailbox message sending and recieving. Created a proxy that extends  IAgent so that Mailbox can support IADS. Mailbox only supports Agents and adding a proxy makes IADS a part of an agent.


Assets/Scripts/IADS/IadsCommsAgent.cs
```diff
@@ -0,0 +1,111 @@
...
+  [SerializeField]
+  private Vector3 _acceleration = Vector3.zero;
+
+  [SerializeField]
+  private Vector3 _accelerationInput = Vector3.zero;
+
+  public HierarchicalAgent HierarchicalAgent {
+    get => _hierarchicalAgent;
+    set => _hierarchicalAgent = value;
+  }
```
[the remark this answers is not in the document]
Joseph0120: Not a bug. In future PR will be resolved

Joseph0120:
@tryuan99 Github tests are failing despite the code working locally. Any reason?

tryuan99:
Two things you should do first:
- Revert all changes to package-lock.json
- Merge in maser into this branch.  `IMessagePayload.cs` is not a new change in this PR

Assets/Scripts/IADS/IadsCommsAgent.cs
```diff
@@ -0,0 +1,118 @@
+using System;
+using UnityEngine;
+
+// IADS Proxy for supporting mailbox message sending and receiving.
+
+public class IadsCommsAgent : MonoBehaviour, IAgent {
```
tryuan99: why not inherit from AgentBase or even DummyAgent?

Assets/Scripts/IADS/IadsCommsAgent.cs
```diff
@@ -0,0 +1,118 @@
+using System;
+using UnityEngine;
+
+// IADS Proxy for supporting mailbox message sending and receiving.
```
tryuan99:
```suggestion
// IADS proxy for supporting mailbox message sending and receiving.
```

Assets/Scripts/IADS/IadsCommsAgent.cs
```diff
@@ -0,0 +1,118 @@
...
+  private HierarchicalAgent _hierarchicalAgent;
+
+  [SerializeField]
+  private Vector3 _velocity = Vector3.zero;
+
+  [SerializeField]
+  private Vector3 _acceleration = Vector3.zero;
+
+  [SerializeField]
+  private Vector3 _accelerationInput = Vector3.zero;
```
tryuan99: add docstrings for these variables

Assets/Tests/EditMode/IadsCommsAgentTests.cs
```diff
@@ -0,0 +1,108 @@
+using NUnit.Framework;
```
tryuan99: do you need to test the IADS comms agent?  It seems to be a dummy agent that has no behavior to test.

tryuan99 (commented):
Why are there 500k lines of deletion?  Are you sure they're all necessary?

Joseph0120:
> Why are there 500k lines of deletion? Are you sure they're all necessary?

@tryuan99 

The deletions are because I deleted the Assets/UnityTechnologies/ParticlePack and I let unity regenerate it. Im not sure if it is the correct procedure. 

At least it works; I think it is because of the version mismatches (new computer).

But it seems like a conflict in the particle pack. I wonder if the particle pack is crucial for this project? Wondering what was the purpose of the particle pack in the beginning when you imported it?

Assets/Scripts/IADS/IadsCommsAgent.cs
Joseph0120: I thought AgentBase is only for physical agents (ie. mass and velocity and other parameters). Goal of IadsCommsAgent is just to add a communication "interface" for the IADS.

tryuan99:
@daniellovell Could you please help review the particle pack changes?

Assets/Scripts/IADS/IadsCommsAgent.cs
tryuan99: Not necessarily, I think.  Right now, you're implementing a lot of pieces of the interface that make it seem like a dummy agent.

tryuan99: Have you addressed this? @Joseph0120

Joseph0120: @tryuan99 I've been thinking about it last week. I think I would propose keeping current because of the following but if you think DummyAgent is better I could still do so :)

Me personally I think because IadsCommsAgent is a comms/mailbox proxy, not a physical simulated agent, IadsCommsAgent intentionally does not support those behaviors and I intentionally not want it to do so; for example it uses SimManager for time and throws on physical/guidance APIs so misuse intentionally fails. I know it is ugly/repetitive but I think the logic make sense

One big thing is that by inheriting AgentBase would make incorrect and unwanted behavior pass through silently (which we don't want). The current design makes accidental use obvious with UnsupportedException; inheritance from DummyAgent would make it look like IADS can sense, pursue, build target models, and etc which we don't want if it in the future has false negatives that passes through (we can't sense it through throwing exceptions).

Also the message layer only needs sender/receiver in Message.cs, but IAgent forces physical properties and behaviors in IAgent.

Assets/Scripts/IADS/IadsCommsAgent.cs
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  public Vector3 AccelerationInput {
+    get => _accelerationInput;
+    set => _accelerationInput = value;
+  }
+
+  public bool IsPursuer => false;
+
+  // Setting ElpasedTime has to be dependent on SimManager to have uniform clock time.
```
tryuan99: typo

Assets/Tests/EditMode/IadsCommsAgentTests.cs.meta
tryuan99, line 1: why is this meta file here with no accompanying `.cs` file?

Assets/Scripts/IADS/IadsCommsAgent.cs
tryuan99: I think the clean solution is one of the following:
- Create a new interface called ICommunicable or something that IADS and Agents inherit from.  I think this is a little bit overkill.
- Use the same composition principle as the hierarchical command structure.  Create a CommsEndpoint or CommsNode class.  Mailbox only operates using these comms nodes.  Then, every agent + IADS has one comms node and listens for activity on the comms node.  This avoids the entire agent hierarchy completely.

Assets/Scripts/Agents/AgentBase.cs
```diff
@@ -3,7 +3,7 @@
 // Base implementation of an agent.
 //
 // See the agent interface for property and method documentation.
-public class AgentBase : MonoBehaviour, IAgent {
+public class AgentBase : MonoBehaviour, IAgent, ICommsNodeOwner {
```
tryuan99: prefer ICommsEndpoint

Assets/Scripts/Communication/CommsNode.cs
```diff
@@ -0,0 +1,52 @@
...
+  public event Action<Message> OnMessageReceived;
+
+  public Configs.AgentType AgentType => _agentOwner?.StaticConfig?.AgentType ?? _agentType;
+
+  public bool IsTerminated {
+    get {
+      if (_isTerminated) {
+        return true;
+      }
+      if (_agentOwner is UnityEngine.Object unityObject && unityObject == null) {
```
tryuan99: this is smelly - why do you need to check the type of a property?

Assets/Scripts/Communication/CommsNode.cs
```diff
@@ -0,0 +1,52 @@
+using System;
+using UnityEngine;
+
+public interface ICommsNodeOwner {
+  CommsNode CommsNode { get; }
+}
+
+// CommsNode is the mailbox-facing endpoint owned by an agent or system such as the IADS.
+public sealed class CommsNode {
+  private readonly IAgent _agentOwner;
```
tryuan99: do you need _agentOwner?  I think you should just have an CommsEndpointType (since IADS is not an agent).

Assets/Scripts/Communication/Message.cs
```diff
@@ -11,13 +11,13 @@ public enum MessageType {
 }
 
 public abstract class Message {
-  public IAgent Sender { get; }
```
tryuan99, line 14: this file looks good!  👍

Assets/Scripts/Communication/CommsNode.cs
```diff
@@ -0,0 +1,52 @@
...
+
+  public CommsNode(Configs.AgentType agentType = Configs.AgentType.InvalidType) {
+    _agentType = agentType;
+  }
+
+  public void Receive(Message message) {
+    if (message == null) {
+      throw new ArgumentNullException(nameof(message));
+    }
+    if (IsTerminated) {
```
tryuan99: I don't think you should check whether it's terminated here.  You just always call the callback, and it's up to the agent to ensure that it unsubscribed from the callback if it died.

Assets/Scripts/Agents/AgentBase.cs
```diff
@@ -109,6 +110,18 @@ public Vector3 AccelerationInput {
...
+  public CommsNode CommsNode { get; private set; }
+
+  public void AttachCommsNode(CommsNode commsNode) {
+    if (commsNode == null) {
+      throw new ArgumentNullException(nameof(commsNode));
+    }
+    if (CommsNode != null && !ReferenceEquals(CommsNode, commsNode)) {
+      throw new InvalidOperationException($"{name} already has a comms node.");
+    }
+    CommsNode = commsNode;
```
tryuan99: overly verbose - it's the responsibility of the comms manager to check this.  the ICommsEndpoint should just have a setter and getter

Assets/Scripts/Communication/CommsNode.cs
```diff
@@ -0,0 +1,44 @@
...
+// in the communication flow. Threats can later be used on the defense side.
+public enum CommsEndpointType {
+  Invalid,
+  Iads,
+  Vessel,
+  ShoreBattery,
+  CarrierInterceptor,
+  MissileInterceptor,
+  FixedWingThreat,
+  RotaryWingThreat,
```
tryuan99: This looks like https://github.com/PisterLab/micromissiles-unity/blob/master/Assets/Proto/Configs/static_config.proto#L6-L17 now.  I think you should reuse that.  In AgentType, I'd add the IADS as index 1 and move the rest up by 1.

Assets/Scripts/Communication/CommsNode.cs
```diff
@@ -0,0 +1,44 @@
...
+
+// CommsNode is the mailbox-facing endpoint owned by an agent or system such as the IADS.
+public sealed class CommsNode {
+  public CommsEndpointType EndpointType { get; }
+
+  private bool _isTerminated;
+
+  public event Action<Message> OnMessageReceived;
+
+  public bool IsTerminated => _isTerminated;
```
tryuan99: why not just a public getter/private setter?

Assets/Scripts/Communication/CommsNode.cs
```diff
@@ -0,0 +1,44 @@
...
+
+  public CommsNode(CommsEndpointType endpointType = CommsEndpointType.Invalid) {
```
tryuan99: no default arguments

Assets/Scripts/IADS/IADS.cs
```diff
@@ -28,6 +28,8 @@ public class IADS : MonoBehaviour {
 
   public IReadOnlyList<IHierarchical> Launchers => _launchers.AsReadOnly();
 
+  public CommsNode CommsNode { get; } = new CommsNode(CommsEndpointType.Iads);
```
tryuan99: no, CommsManager should listen for `OnSimulationStarted` and `OnSimulationEnded` and allocate/destroy a comms node to IADS in those callbacks

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+// CommsManager owns the lifecycle of interceptor comms nodes and keeps that lifecycle tied to
+// simulation events rather than Unity object construction and destruction.
+public class CommsManager : MonoBehaviour {
+  public static CommsManager Instance { get; private set; }
+
+  private readonly Dictionary<IAgent, CommsNode> _nodesByAgent =
+      new Dictionary<IAgent, CommsNode>();
+
+  // Ensure the comms manager exists before scene objects begin their normal lifecycle.
+  [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
```
tryuan99: There's a way in Unity to define the priority of these managers, so that they start up in a deterministic order.  You should do that instead to be consistent

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+    if (Instance != null) {
+      return;
+    }
+
+    var managerObject = new GameObject("CommsManager");
+    DontDestroyOnLoad(managerObject);
+    managerObject.AddComponent<CommsManager>();
+  }
+
+  // Enforce a single persistent comms manager instance across scene loads.
```
tryuan99: no need for this comment

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+    }
+    Instance = this;
+    DontDestroyOnLoad(gameObject);
+  }
+
+  // Wait for the simulation manager, then subscribe to interceptor lifecycle events and catch up
+  // any interceptors that already exist.
+  private IEnumerator Start() {
+    while (SimManager.Instance == null) {
+      yield return null;
```
tryuan99: no need for this if you define the priorities of the managers

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    SimManager.Instance.OnSimulationEnded += RegisterSimulationEnded;
+    SimManager.Instance.OnNewInterceptor += RegisterNewInterceptor;
+
+    foreach (var agent in SimManager.Instance.Interceptors) {
```
tryuan99: no, comms manager should be there before interceptors get added.  you then listen for RegisterNewInterceptor to add a comms node to each newly created interceptor

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        RegisterNewInterceptor(interceptor);
+      }
+    }
+  }
+
+  // Unsubscribe from simulation events and terminate any endpoints still owned by this manager.
+  private void OnDestroy() {
+    if (SimManager.Instance != null) {
+      SimManager.Instance.OnSimulationEnded -= RegisterSimulationEnded;
+      SimManager.Instance.OnNewInterceptor -= RegisterNewInterceptor;
```
tryuan99: no need

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    agent.OnTerminated -= RegisterAgentTerminated;
+    commsNode.Terminate();
+    _nodesByAgent.Remove(agent);
+  }
+
+  // Terminate every tracked comms node when the simulation ends or the manager is torn down.
+  private void RegisterSimulationEnded() {
+    foreach (var agentNodePair in _nodesByAgent) {
```
tryuan99: is this necessary?  the agent will die anyway.  why not just clear _nodesByAgent?

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -0,0 +1,112 @@
...
+      agentNodePair.Key.OnTerminated -= RegisterAgentTerminated;
+      agentNodePair.Value.Terminate();
+    }
+    _nodesByAgent.Clear();
+  }
+
+  private static CommsEndpointType GetEndpointType(IInterceptor interceptor) {
+    return interceptor.StaticConfig?.AgentType switch {
```
tryuan99: if you use AgentType, you don't need this map :)

Pisterlab_Micromissile.slnx
```diff
@@ -0,0 +1,7 @@
+﻿<Solution>
```
tryuan99: what is this file?

.github/workflows/build.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          - ubuntu-latest
+          - ubuntu-22.04
```
tryuan99: check if this is necessary

Assets/UniversalRenderPipelineGlobalSettings.asset
```diff
@@ -126,16 +124,18 @@ MonoBehaviour:
         m_CoreBlitPS: {fileID: 4800000, guid: 93446b5c5339d4f00b85c159e1159b7c, type: 3}
         m_CoreBlitColorAndDepthPS: {fileID: 4800000, guid: d104b2fc1ca6445babb8e90b0758136b, type: 3}
         m_SamplingPS: {fileID: 4800000, guid: 04c410c9937594faa893a11dceb85f7e, type: 3}
-        m_TerrainDetailLit: {fileID: 4800000, guid: f6783ab646d374f94b199774402a5144, type: 3}
-        m_TerrainDetailGrassBillboard: {fileID: 4800000, guid: 29868e73b638e48ca99a19ea58c48d90, type: 3}
-        m_TerrainDetailGrass: {fileID: 4800000, guid: e507fdfead5ca47e8b9a768b51c291a1, type: 3}
+        m_TerrainDetailLit: {fileID: 0}
```
tryuan99: fileID = 0 is not supposed to happen.  Can you regenerate this file?

Pisterlab_Micromissile.slnx
tryuan99: same with this file.  I'd try deleting and then running a simulation to see what is automatically re-generated

tryuan99 (commented):
Just two comments on generated files.  Everything lgtm

PR #217, opened 2026-04-01
[Communication] PR5: IADS.cs now supports Mailbox

Joseph0120:
Modified Files:

IADS.cs -> IADS can now send, receive, handle messages, and subscriptions go through mailbox instead of straight to agents.

### DIFF
--- a/Assets/Scripts/IADS/IADS.cs
+++ b/Assets/Scripts/IADS/IADS.cs
@@ -57,9 +57,10 @@ public class IADS : MonoBehaviour {
       // Just in case the simulation started before IADS subscribed to events.
       foreach (var agent in SimManager.Instance.Interceptors) {
         if (agent is IInterceptor interceptor) {
-          RegisterNewAsset(interceptor);
-          if (interceptor is LauncherBase) {
-            RegisterNewLauncher(interceptor);
+          if (interceptor is LauncherBase launcher) {
+            RegisterNewLauncher(launcher);
+          } else {
+            RegisterNewAsset(interceptor);
           }
         }
       }
@@ -216,7 +217,8 @@ public class IADS : MonoBehaviour {
   }
 
   private void AssignSubInterceptor(IInterceptor subInterceptor) {
-    if (subInterceptor == null || subInterceptor.CapacityRemaining <= 0) {
+    if (subInterceptor == null || subInterceptor.HierarchicalAgent == null ||
+        subInterceptor.CapacityRemaining <= 0) {
       return;
     }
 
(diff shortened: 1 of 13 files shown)


Assets/Scripts/IADS/IADS.cs
```diff
@@ -11,16 +11,21 @@ public class IADS : MonoBehaviour {
...
   private Coroutine _hierarchyCoroutine;
 
   // List of threats waiting to be incorporated into the hierarchy.
-  private List<IHierarchical> _newThreats = new List<IHierarchical>();
+  private readonly List<IHierarchical> _newThreats = new List<IHierarchical>();
+
+  // Routes mailbox traffic through the IADS proxy agent.
+  private IadsCommsAgent _commsAgent;
+  private Mailbox _mailboxInstance;
+  private bool _mailboxSubscribed = false;
```
tryuan99, line 28: ditto from previous PR

Assets/Scripts/IADS/IADS.cs
```diff
@@ -31,28 +36,78 @@ public class IADS : MonoBehaviour {
...
     }
   }
 
   private void Start() {
-    SimManager.Instance.OnSimulationStarted += RegisterSimulationStarted;
-    SimManager.Instance.OnSimulationEnded += RegisterSimulationEnded;
-    SimManager.Instance.OnNewAsset += RegisterNewAsset;
-    SimManager.Instance.OnNewLauncher += RegisterNewLauncher;
-    SimManager.Instance.OnNewThreat += RegisterNewThreat;
+    if (SimManager.Instance != null) {
```
tryuan99, line 50: why do you need this check now?

Assets/Scripts/IADS/IADS.cs
```diff
@@ -31,28 +36,78 @@ public class IADS : MonoBehaviour {
...
+      SimManager.Instance.OnSimulationStarted += RegisterSimulationStarted;
+      SimManager.Instance.OnSimulationEnded += RegisterSimulationEnded;
+      SimManager.Instance.OnNewAsset += RegisterNewAsset;
+      SimManager.Instance.OnNewLauncher += RegisterNewLauncher;
+      SimManager.Instance.OnNewThreat += RegisterNewThreat;
+
+      // Just in case the simulation started before IADS subscribed to events.
```
tryuan99, line 57: will this ever happen?  no need to add if it will never happen

Assets/Scripts/IADS/IADS.cs
```diff
@@ -31,28 +36,78 @@ public class IADS : MonoBehaviour {
...
+        RegisterSimulationStarted();
+      }
+    }
+
+    TrySubscribeMailbox();
+  }
+
+  private void LateUpdate() {
+    if (!_mailboxSubscribed) {
+      TrySubscribeMailbox();
```
tryuan99, line 82: only call it once.  if there's race condition, this doesn't fix the problem

tryuan99 (commented):
Please check every change in this PR and ensure that it's absolutely necessary.  No need for extra checks and overly preventative code that makes readability worse.  If a problem happens, it'll be logged and we can fix then :)

PR #229, opened 2026-07-06
[Unity] Upgrade to Unity 6.3 LTS

tryuan99:
This PR bumps the Unity editor version to Unity 6.3 (6000.3.19f1).  There is a warning that the InputManager is soon to be deprecated, so I have ported over some code to the newer Input System: https://docs.unity3d.com/Packages/com.unity.inputsystem@1.19/manual/Installation.html.


daniellovell (approved):
Seems to work ok for me on the new Unity 6.3, but I get these suspicious warnings. Do you see these?

Image

tryuan99:
Those warnings were there before: The CVN has no sensor and controller, and the DDG has no controller.  I think we should give the CVN an ideal sensor (it never gets assigned a target anyway, but it might be useful for evasion (at 5 knots lol)).  For controller, the direct waypoint controller is the correct choice.

PR #231, opened 2026-07-07
[CI] Fix Bazel cache

tryuan99:
This PR fixes the Bazel cache, so that it is used for builds and subsequent tests, both for the plugins and for the Unity library.  The caches are also now scoped by OS, and external Bazel dependencies are cached as well.  Finally, this PR bumps the versions of many CI tools.

Building the plugins for Windows now is pretty quick thanks to the Bazel cache:
image


daniellovell (approved):
W

PR #237, opened 2026-07-30
[Communication v2] PR1: Tied Mailbox to CommsManager

Joseph0120:
This PR adds the mailbox to the comms manager to enable agents to send messages to each other with a set latency.  The mailbox is polled every fixed update to check for due messages to be delivered.

### DIFF
--- a/Assets/Scripts/Communication/Mailbox.cs
+++ b/Assets/Scripts/Communication/Mailbox.cs
@@ -1,93 +1,72 @@
 using System;
 using UnityEngine;
 
-// The mailbox handles the communication latencies between agent to agent communication.
-//
-// Mailbox.cs owns the message queue and releases messages to deliver after the set latency in
-// communication_config.
+// The mailbox handles the communication latencies for agent-to-agent communication. The mailbox.cs
+// owns the message queue and releases messages to be delivered after a set latency.
 public class Mailbox {
-  // OnMessageReceived is for logging purposes.
-  public event Action<Message> OnMessageReceived;
-
-  private static readonly Configs.LinkConfig _fallbackLinkConfig = new Configs.LinkConfig {
+  private static readonly Configs.LinkConfig _defaultLinkConfig = new Configs.LinkConfig {
     PacketDeliveryRatio = 1f,
   };
 
+  // Message queue.
   private readonly PriorityQueue<PendingMessage> _messageQueue =
       new PriorityQueue<PendingMessage>();
 
-  // SendMessage checks and enqueues a message to be sent to the receiver. The message will be
-  // delivered after the latency has been applied.
-  public void SendMessage(Message message) {
+  // Enqueue a message to be delivered after a set latency.
+  public void Send(Message message) {
     if (message == null) {
       return;
     }
-    CommsManager commsManager = CommsManager.Instance;
-    // Guard, Sender and Receiver must be valid.
-    if (commsManager == null || !commsManager.ContainsNode(message.Receiver)) {
-      return;
+    if (CommsManager.Instance.ContainsNode(message.Receiver)) {
+      Configs.LinkConfig config = GetLinkConfig(message);
+
+      // TODO(Joseph0120): Set the packet delivery ratio to config.PacketDeliveryRatio.
+      float packetDeliveryRatio = Mathf.Clamp01(1);
+      if (UnityEngine.Random.value >= packetDeliveryRatio) {
+        return;
+      }
+
+      float latency = config.LatencySeconds;
+      float jitter = Utilities.SampleStandardNormal() * config.LatencyStdSeconds;
+      float totalLatency = Math.Max(0f, latency + jitter);
+      float deliverAt = SimManager.Instance.ElapsedTime + totalLatency;
+
+      var pendingMessage = new PendingMessage(message, deliverAt);
+      _messageQueue.Enqueue(pendingMessage, pendingMessage.DeliverAt);
     }
-    EnqueueMessage(message);
   }
 
-  // EnqueueMessage adds a message to the message queue with the appropriate latency applied.
-  private void EnqueueMessage(Message message) {
-    Configs.LinkConfig config = GetLinkConfig(message);
-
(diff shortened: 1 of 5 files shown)


Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,104 @@
+using System;
+using UnityEngine;
+
+public class Mailbox {
+  public static Mailbox Instance { get; private set; }
```
tryuan99: I don't think you need static instance.

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,104 @@
...
+    }
+    EnqueueMessage(message);
+  }
+
+  private void EnqueueMessage(Message message) {
+    Configs.LinkConfig config = GetLinkConfig(message);
+
+    // TODO (Joseph): This packet delivery ratio is a temporary solution. Fundamental
+    // firing/communication logic needs to change.
+    float packetDeliveryRatio = ToValidDeliveryRatio(config.PacketDeliveryRatio);
```
tryuan99: assume configs are correct and valid.

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,104 @@
...
+    float deliverAt = (SimManager.Instance?.ElapsedTime ?? 0f) + totalLatency;
+
+    // pendingMessage is a wrapper that includes the message and deliverAt.
+    var pendingMessage = new PendingMessage(message, deliverAt);
+    _messageQueue.Enqueue(pendingMessage, pendingMessage.DeliverAt);
+  }
+
+  private Configs.LinkConfig GetLinkConfig(Message message) {
+    Configs.CommunicationConfig communicationConfig =
+        SimManager.Instance?.SimulationConfig?.CommunicationConfig;
```
tryuan99: SimManager.Instance should always exist

Pisterlab_Micromissile.slnx
```diff
@@ -0,0 +1,9 @@
+﻿<Solution>
```
tryuan99: remove this file

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,91 @@
+using System;
+using UnityEngine;
```
tryuan99, line 2: please add docstrings here and at least one comment for each private variable (hand-written!)

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,91 @@
...
+
+public class Mailbox {
+  // OnMessageReceived is for logging purposes.
```
tryuan99: no need for this comment because I think this can be used for other things too like statistics

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,91 @@
...
+    EnqueueMessage(message);
+  }
+
+  private void EnqueueMessage(Message message) {
+    Configs.LinkConfig config = GetLinkConfig(message);
+
+    // TODO (Joseph): This packet delivery ratio is a temporary solution. Fundamental
+    // firing/communication logic needs to change.
+    float packetDeliveryRatio = config.PacketDeliveryRatio;
+    if (packetDeliveryRatio <= 0f ||
```
tryuan99: I would just clamp it to 01

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,91 @@
...
+      }
+    }
+    return communicationConfig.LinkConfig ?? _fallbackLinkConfig;
+  }
+
+  public void UpdateMailbox() {
+    CommsManager commsManager = CommsManager.Instance;
+    SimManager simManager = SimManager.Instance;
+    if (commsManager == null || simManager == null || !simManager.IsRunning) {
+      return;
```
tryuan99: assume comms manager and sim manager must be running, and they should be or something is wrong

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    while (!_messageQueue.IsEmpty() && _messageQueue.Peek().DeliverAt <= simManager.ElapsedTime) {
+      var pendingMessage = _messageQueue.Dequeue();
+      // Guard, Sender and Receiver must be valid.
+      if (!commsManager.ContainsNode(pendingMessage.Sender) ||
```
tryuan99: sender may be dead, that's ok

Assets/Scripts/Communication/Mailbox.cs
```diff
@@ -0,0 +1,91 @@
...
+          !commsManager.ContainsNode(pendingMessage.Receiver)) {
+        continue;
+      }
+      pendingMessage.Receiver.Receive(pendingMessage.Message);
+      // TODO (Joseph): OnMessageReceived for logging.
+      OnMessageReceived?.Invoke(pendingMessage.Message);
+    }
+  }
+
+  public void ClearPendingMessageQueue() {
```
tryuan99: if not used in this PR, remove

Assets/Scripts/Communication/Mailbox.cs
tryuan99: Actually, I don't think we need this event.  Mailbox shouldn't be responsible for events.  If you want this, I'd suggest giving a new event to CommsManager called OnMessageDelivered or something

tryuan99 (commented):
a few comments and otherwise it's good

Assets/Scripts/Communication/Mailbox.cs
Joseph0120: I just deleted the event. If we want logging we can add it in the future.

Assets/Scripts/Communication/Mailbox.cs
Joseph0120: Yes everything is hand written :)

tryuan99 (approved):

PR #246, opened 2026-09-22
Added Cooldown and Packet Drop Behaviour

Joseph0120:
Added cooldown time in Communication_config.proto. Example shown in 4_swarms_4_ucav.pbtxt.

Cooldown time is the amount of buffer time that SAME messages have to wait before resending again. 

Packet Drop Ratio is corrected implemented in this PR. Packet Drop is implemented after passing the cooldown condition but before enqueueing into the mailbox. Verified with .csv and unit tests that cooldown and PDR behaves as intended.


Assets/Proto/Configs/communication_config.proto
```diff
@@ -33,4 +33,9 @@ message CommunicationConfig {
 
   // Per-pair overrides. Omitted pairs fall back to link_config.
   repeated LinkOverride link_overrides = 2;
+
+  // Minimum time between identical messages from the same sender to the same
+  // receiver. Changed messages bypass the cooldown. Must be non-negative; zero
+  // disables it.
+  float cooldown_seconds = 3;
```
tryuan99, line 40: well not really - the minimum is the step time of the simulation

Assets/Scripts/Managers/CommsManager.cs
```diff
@@ -46,15 +46,18 @@ private void RegisterSimulationEnded() {
...
+    if (commsNode == null) {
+      commsNode = new CommsNode(agent.StaticConfig.AgentType);
+      agent.CommsNode = commsNode;
     }
 
-    var commsNode = new CommsNode(agent.StaticConfig.AgentType);
-    agent.CommsNode = commsNode;
-    agent.OnTerminated +=
-        _ => _nodes.Remove(commsNode);
-    _nodes.Add(commsNode);
```
tryuan99, line 58: what's the reason for this change?  seems overly defensive

Assets/Scripts/Managers/SimManager.cs
```diff
@@ -19,7 +19,7 @@ public class SimManager : MonoBehaviour {
   public event Action<IThreat> OnNewThreat;
 
   // Default simulation configuration file.
-  private const string _defaultSimulationConfigFile = "7_quadcopters.pbtxt";
+  private const string _defaultSimulationConfigFile = "4_swarms_4_ucav.pbtxt";
```
tryuan99, line 22: keep the default unchanged

tryuan99 (changes requested):
The mailbox should be a dumb interface.  It's a nice abstraction currently, let's not bloat it with unrelated functionality.

It should not need to worry about re-sending messages.  A better approach is that each agent should have some state machine about its target status like 1) target acquired, 2) no target, 3) requested target.

In state 3), the agent should then use a timer to wait on a response.  If no response occurs within some time, then re-request.

Assets/Proto/Configs/communication_config.proto
Joseph0120: I understand that, but it can’t be zero?

should it be divisible by the step time then? But that’s hard for the user

tryuan99: I meant the comment is misleading

